PR #3, opened 2014-10-26
export option for PBCore 2.0

dericed:
—output=PBCore2


dericed:
merged via sourceforge not ago

PR #4, opened 2014-12-22
Donotskip

kodabb:
I did some analysis on the missing Width and Height properties not being reported for some quicktime files.
 
The problem seems to be located in the bitstream functions which check the size of the buffer before skipping it. However when they trigger a possible overrun, they return immediately (not increasing the buffer pointer) and raise a flag that the whole atom is "untrusted". Consequently the parsed values (even the correct ones, such as width and height) do not get propagated to the gui and just skipped.

What is available here is a "workaround" set that tries to use the 'get'-functions (rather than the 'skip'-) bitstream functions, which do not trigger this problem. Alternatively one could avoid touching those functions completely, and always write Width and Height, but I am not sure of unforeseen problems.


JeromeMartinez:
"mov: parse the color table only when file is palettized" accepted
"mov: completely skip the 'encoder' tag" rejected (another fix was used)

PR #9, opened 2015-08-03
mediatrace updates

dericed:
This line https://github.com/dericed/MediaInfoLib/blob/master/Source/MediaInfo/File__Analyze.cpp#L2580 still needs an update to replace <data> with <block> rather than appending an 's' as it currently does


JeromeMartinez:
> still needs an update to replace with rather than appending an 's' as it currently does

                                                                            {
                                                                               Element[Element_Level].ToShow.Details.erase(item_Pos+ToFind.size()-4, 4);
                                                                               Element[Element_Level].ToShow.Details.insert(item_Pos+ToFind.size()-4, __T("block"));
                                                                            }

PR #14, opened 2015-08-12
Video/FFV1: use the quant_table_index for the context count by plane

tribouille:
Signed-off-by: Florent Tribouilloy <florent@mediaarea.net>


JeromeMartinez:
.py patch is out of topic.

tribouille:
Yep, sorry.

PR #20, opened 2015-11-09
Matroska Statistics Tags Fix

ndjamena:
It shouldn't be that hard to modify to simply remove any trace of statistics tags from display.

I've modified the default track duration method to only be applied if the duration tag isn't present in case you want to keep it, but I've commented it out.  Of course, if you're going to do a full pass-through of the file that might be good enough to override the tag.

If you think there might be any lone BPS tags out there somewhere that you don't want, you might add some extra clean-up code in the stream finish section to deal with it, or with any other lonely stats tags.

I've left it to display any unknown Stat Tags (or any with an incorrect security tag) with an asterisk before them, if you just want to remove them completely that shouldn't be hard.  Or you could go the other way and display the security tags when they don't match, just to give the user some idea as to why.

Is it necessary to add code to better scrutinise the values being given?  If BPS + Duration is more than the size of the file or the duration is one hour and eighty eight minutes?  ...Or would that be overkill?

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mk.cpp
+++ b/Source/MediaInfo/Multiple/File_Mk.cpp
@@ -216,7 +216,7 @@ void File_Mk::Streams_Finish()
 	//Ztring Duration_Temp;
 	bool Tags_Verified=false; 
 	Ztring TagsList=Retrieve(StreamKind_Last, StreamPos_Last, "_STATISTICS_TAGS");
-	if (!TagsList.empty())
+	if (TagsList.size())
 	{
 		Clear(StreamKind_Last, StreamPos_Last, "_STATISTICS_TAGS");
 		if (Retrieve(StreamKind_Last, StreamPos_Last, "_STATISTICS_WRITING_APP")==(Retrieve(Stream_General, 0, "Encoded_Application")) &&
--- a/Source/MediaInfo/Multiple/File_Mk.cpp
+++ b/Source/MediaInfo/Multiple/File_Mk.cpp
@@ -3484,8 +3484,8 @@ void File_Mk::CodecID_Manage()
     else if (Format==__T("FFV1"))
     {
         Stream[TrackNumber].Parser=new File_Ffv1;
-        ((File_Ffv1*)Stream[TrackNumber].Parser)->Width=Retrieve(Stream_Video, StreamPos_Last, Video_Width).To_int64u();
-        ((File_Ffv1*)Stream[TrackNumber].Parser)->Height=Retrieve(Stream_Video, StreamPos_Last, Video_Height).To_int64u();
+        ((File_Ffv1*)Stream[TrackNumber].Parser)->Width=Retrieve(Stream_Video, StreamPos_Last, Video_Width).To_int32u();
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
Why "Statistics Tags Verified"?
I have a file with "Lavf55.19.104" writing application and "mkvmerge v7.1.0" _STATISTICS_WRITING_APP and it looks good (more or less, there is a delat of few seconds between Matroska duration and tags, I did not check which one is right).

Any reason for not trusting data when both values are not same?

ndjamena:
<Why "Statistics Tags Verified"?>
It's preliminary, in case there's unknown statistics tags, it lists them as is with an asterisk to show that they're part of that group and that they're verified (or not).

It's all unnecessary, I just don't really know what you want.  If you want them all gone then they're gone.

<Any reason for not trusting data when both values are not same?>

...because they may have been re-encoded (or trimmed) and in that case all the statistics will be wrong.  If you think being able to see "bad" tags will be useful you can keep them there, it's up to you really, I can just hide everything if you like.

JeromeMartinez:
> ...because they may have been re-encoded (or trimmed) and in that case all the statistics will be wrong.

Good point. So let's keep this filter for the moment.

ndjamena:
Actually I should probably hide "*Statistics Tags Verified" when there are no stray tags around, but I'm being cheap with variables.

JeromeMartinez:
I plan some modifications like hiding this field, right.
(also removal of "*", still thinking about "right" field names)

ndjamena:
I need to figure out how to access the Matroska bitrate from the DTS parser...  I assume it works that way since DTS bitrate overrides the Matroska one...

Would the DTS-MA bitrate be the combined Core + HD extensions or just the extensions?

JeromeMartinez:
> I need to figure out how to access the Matroska bitrate from the DTS parser.

It is not easy for someone who does not know well MI code (it is during "::Merge()" call).
Here, it is a bit some string manipulation, "Unknown / xxx" must be changed to "yyy / xxx" with yyy the Matroska bitrate and xxx the Core bitrate.

if you are motivated, you need to get the bitrate before calling :
Merge(*Temp->second.Parser, Temp->second.StreamKind, 0, Temp->second.StreamPos);
then get the "new" bitrate after the "merge" and manipulate the string.

> Would the DTS-MA bitrate be the combined Core + HD extensions or just the extensions?

with yyy / xxx, my idea is to have yyy the total bitrate (Core + HD) and xxx the Core bitrate.

***

Please provide a DTS-MA MKV with tags and I'll handle this case.

BTW, please also contact me for some paperwork.

ndjamena:
http://www.mediafire.com/watch/mlsax4rpqjodrul/00001.m2ts_%281%29.mkv

Grrr, I forgot.  I have it in an email that Mosu decreed that any tag whose name begins with an underscore '_' is NOT user viewable and should be hidden.  I was going to code that in but forgot.

Duration / Frames = Framerate, if it doesn't match the calculated framerate then it must be VFR...  UNLESS of course it's a broken file with "gaps"...

I figure what you need to do is if you find a video whose tag frame rate doesn't match the calculated one, you mark it as broken.  Audio tends to have fixed frame rates, if Duration / Frames for all the audio tracks gives the correct frame rate then it's most likely NOT the file itself that's broken and you can then tag the video as VFR with at least some confidence.

PR #24, opened 2015-11-19
Set VFR if framerate of first frames not equal to framerate implied in Matroska Statistics Tags.

ndjamena:
Set VFR if framerate of first frames not equal to framerate implied in Statistics Tags.

(TimecodeScale*0.0000000010021) is the margin of error needed to detect 59.880 fps as being equal to 60/1.001 fps with a timescale of 1000000ns.  Tweak it as you like.

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mk.cpp
+++ b/Source/MediaInfo/Multiple/File_Mk.cpp
@@ -239,6 +239,7 @@ void File_Mk::Streams_Finish()
     
         //Tags
         //Ztring Duration_Temp;
+        float32 FrameRate_FromTags = 0.0;
         Ztring TagsList=Retrieve(StreamKind_Last, StreamPos_Last, "_STATISTICS_TAGS");
         if (TagsList.size())
         {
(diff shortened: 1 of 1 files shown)


ndjamena:
If I remove this:

```

                    if (Temp->second.TimeCodes.size()>30+FramesToAdd)
                        Time=(float)(Temp->second.TimeCodes[30+FramesToAdd]-Temp->second.TimeCodes[0])/30; //30 frames for handling 30 fps rounding problems
                    else

```

And set time like this instead:

```
                    Time=(float)(Temp->second.TimeCodes[Temp->second.TimeCodes.size()-1]-Temp->second.TimeCodes[0])/(Temp->second.TimeCodes.size()-1);
                    if (Time*0.9<Temp->second.TrackDefaultDuration && Time*1.1>Temp->second.TrackDefaultDuration)
                        Time=(float)Temp->second.TrackDefaultDuration/TimecodeScale; //TrackDefaultDuration is maybe more precise than the time code
```

I can avoid the whole 59.880fps thing completely.

I just don't know what it means by "30 frames for handling 30 fps rounding problems".  What problem is it fixing and how?

ndjamena:
I've created a file with 30fps and another with 30/1.001 fps, stripped them of tags and each one has been detected perfectly as their respective frame rates using my code... so unless "30fps rounding problems" means something else then the current line should go, it's just screwing up 60/1.001 for no purpose.

ndjamena:
I haven't changed the code yet.  I don't know what that line actually does and I won't propose a change until I know it's not necessary.

PR #33, opened 2015-11-28
x Fixed compilation issues with some #defines

kasper93:
Compilation currently fail when MEDIAINFO_XML_YES and/or MEDIAINFO_ADVANCED is not defined.


JeromeMartinez:
thanks!

PR #40, opened 2015-12-02
Cover data in Matroska

maxpozdeev:
I rewrote the code of getting cover data in matroska. Please review.

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mk.cpp
+++ b/Source/MediaInfo/Multiple/File_Mk.cpp
@@ -91,8 +91,8 @@
 #include "MediaInfo/MediaInfo_Config_MediaInfo.h"
 #include <cstring>
 #include <algorithm>
+#include "base64.h"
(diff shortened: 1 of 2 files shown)


JeromeMartinez:
the PR has some issues:
- 3 "Merge remote-tracking branch"
- code then fix, it is better to revert the patch then new patch.
- it removes the "LIST" item and it replaces it with an ATOM, but "LISt" is there because I have some files with huge attachments and the file is rejected in that case because MediaInfo considers it is not expected so risky. There is also no limitation in the size of the attachment so someone could crash MediaInfo with a Matroska file having an attachment of e.g. 1 GB (if I remove the security about the maximum acceptable size of an ATOM)

so I slightly modified the patch and committed it https://github.com/MediaArea/MediaInfoLib/commit/0b234f787f15aa4652dfc555856db2cc2b389841 (you are still the author of the patch).

PR #47, opened 2015-12-05
Mac OS X support for the Python interface

miguelgrinberg:
I've made the following changes:
- Argument lists for most calls were set in a "argtype" attribute which
  is wrong. The correct attribute name is "argtypes".
- Added support to load the Mac OS X dynamic library.
- Modified the Python 3 version of the library to also be compatible
  with Python 2, so that a single file can be used on both.

### DIFF
--- a/Source/MediaInfoDLL/MediaInfoDLL.py
+++ b/Source/MediaInfoDLL/MediaInfoDLL.py
@@ -13,14 +13,19 @@
 # Converted to python module by Petr Kaderabek
 # Modifications by Jerome Martinez
 # Python 3 update by Jerome Martinez
+# Mac OSX support and ctypes fixes by Miguel Grinberg
 #
 # +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++*/
 
 import os
+import sys
 from ctypes import *
 if os.name == "nt" or os.name == "dos" or os.name == "os2" or os.name == "ce":
     MediaInfoDLL_Handler = windll.MediaInfo
     MustUseAnsi = 0
+elif sys.platform == "darwin":
+    MediaInfoDLL_Handler = CDLL("libmediainfo.0.dylib")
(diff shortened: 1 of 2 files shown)


JeromeMartinez:
- "argtype" wrong? I understood it was automaticly detected. Anyway it is better to be explicit.
- Great!
- So if I replace MediaInfoDLL.py by MediaInfoDLL3.py and HowToUse_Dll.py by HowToUse_Dll3.py, it is OK (both Python 2 and Python 3 can use any of the files), without regression, right? I tested myself but I prefer to be sure I do nothing wrong.

miguelgrinberg:
> "argtype" wrong? I understood it was automaticly detected. Anyway it is better to be explicit.

Yeah, on the Mac "argtype" does not give any errors, but ctypes ignores it and goes with the default of all arguments being 32-bit ints. Things fail pretty epicly :)

> So if I replace MediaInfoDLL.py by MediaInfoDLL3.py and HowToUse_Dll.py by HowToUse_Dll3.py

Yes, that was my intention. I did not remove the Python 2 version to not break people that is already using it (such as the python-mediainfodll ubuntu pkg), but you can now use the version 3 code on both versions of Python.

Thanks for the prompt handling!

JeromeMartinez:
> Things fail pretty epicly.

Ouch. Weird that it works well on Linux and Window.
hum... actually, doing the diff of MediaInfoDLL.py and MediaInfoDLL3.py, I see a comment:
"#We should use -1 the same way, but it fails on 64-bit"
so it was working but not so well :).

> I did not remove the Python 2 version to not break people that is already using it

The issue is definitely legacy programs, I don't see how I could avoid to keep duplicate content.
Please review https://github.com/MediaArea/MediaInfoLib/pull/49, is it OK from your point of view? (I tested on Windows but I prefer someone else double-checks)

PR #49, opened 2015-12-05
Python 2/3 files merge

JeromeMartinez:
HowToUse_Dll.py and HowToUse_Dll3.py are same
MediaInfoDLL.py and MediaInfoDLL3.py are same
but all files are kept in order to not break programs calling them.

### DIFF
--- a/Source/Example/HowToUse_Dll.py
+++ b/Source/Example/HowToUse_Dll.py
@@ -6,21 +6,15 @@
 
 #+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 #
-# Python (Windows)  example
+# Python example
 #
-# To make this example working, you must put MediaInfo.Dll and test.avi
-# in the same folder
+# To make this example working, you must put MediaInfo.Dll, MediaInfoDLL.py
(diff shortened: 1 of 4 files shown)


JeromeMartinez:
Idea from https://github.com/MediaArea/MediaInfoLib/pull/47.
@miguelgrinberg would you mind to review this patch?

miguelgrinberg:
Looks good to me.

But note that one thing I'm not 100% sure is if this 2/3 version of the code will run on older Python versions. I tested it on Python 2.7, and I verified that the changes I've made are also fine for 2.6. Older versions may or may not work. If this is a concern for you, then maybe merging is not a good idea.

I personally haven't used any Python release older than 2.7 in a very long time, so I'm :+1: with this change.

JeromeMartinez:
[ctypes](https://docs.python.org/2/library/ctypes.html) and CDLL are expected to be present in 2.5 but impossible to run the example:
ImportError: No module named ctypes
both before and after your patch.

And 2.5 is "End Of Life" (even RHEL 6 has already Python 2.6 :) ), so that's OK if it runs on 2.6 (tested also on Windows, also OK).

PR #71, opened 2015-12-21
MakeMKV Tags and (ever so slightly) less faulty frame rate detections

ndjamena:
MakeMKV has added it's own tags into the fray.

I have a TV episode at NTSC film frame rate with exactly 29821 frames...
Do I need to show the math?

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mk.cpp
+++ b/Source/MediaInfo/Multiple/File_Mk.cpp
@@ -323,9 +323,32 @@ void File_Mk::Streams_Finish()
                             else if (TempTag==__T("NUMBER_OF_BYTES"))
                             {
                                 if (Tags_Verified)
-                                    { Fill(StreamKind_Last, StreamPos_Last, Fill_Parameter(StreamKind_Last, Generic_StreamSize), TagValue, true); Set=true; }
+                                {
+                                    Fill(StreamKind_Last, StreamPos_Last, Fill_Parameter(StreamKind_Last, Generic_StreamSize), TagValue, true);
+                                    Set=true;
+                                }
                                 else
                                     TempTag="StreamSize";
+                            }
+                            else if (TempTag==__T("NUMBER_OF_BYTES_UNCOMPRESSED"))
+                            {
+                                if (Tags_Verified)
+                                {
+                                    Fill(StreamKind_Last, StreamPos_Last, "Stream Size (Uncompressed)", TagValue, true);
+                                    Set=true;
+                                }
+                                else
+                                    TempTag="Stream Size (Uncompressed)";
+                            }
+                            else if (TempTag==__T("SOURCE_ID"))
+                            {
+                                if (Tags_Verified)
+                                {
+                                    Fill(StreamKind_Last, StreamPos_Last, "Source ID", TagValue, true);
+                                    Set=true;
+                                }
+                                else
+                                    TempTag="Source ID";
                             }
                             if (!Set)
                             {
@@ -350,20 +373,24 @@ void File_Mk::Streams_Finish()
                     // Checking 1.001 frame rates, Statistics_Duration is has often only a 1 ms precision so we test between -1ms and +1ms
                     float64 Duration_1001 = Statistics_FrameCount / float64_int64s(FrameRate_FromTags) * 1.001000;
                     float64 Duration_1000 = Statistics_FrameCount / float64_int64s(FrameRate_FromTags) * 1.001001;
+                    bool CanBe1001 = false;
+                    bool CanBe1000 = false;
+                    if (abs((Duration_1000 - Duration_1001) * 10000) >= 15)
+                    {
+                        Ztring DurationS; DurationS.From_Number(Statistics_Duration, 3);
+                        Ztring DurationS_1001; DurationS_1001.From_Number(Duration_1001, 3);
+                        Ztring DurationS_1000; DurationS_1000.From_Number(Duration_1000, 3);
 
-                    Ztring DurationS; DurationS.From_Number(Statistics_Duration, 3);
-                    Ztring DurationS_1001; DurationS_1001.From_Number(Duration_1001, 3);
-                    Ztring DurationS_1000; DurationS_1000.From_Number(Duration_1000, 3);
-
-                    bool CanBe1001=DurationS==DurationS_1001?true:false;
-                    bool CanBe1000=DurationS==DurationS_1000?true:false;
-                    if (CanBe1001 && !CanBe1000)
-                        FrameRate_FromTags = float64_int64s(FrameRate_FromTags) / 1.001;
-                    if (CanBe1000 && !CanBe1001)
-                        FrameRate_FromTags = float64_int64s(FrameRate_FromTags) / 1.001001;
+                        CanBe1001=DurationS==DurationS_1001?true:false;
+                        CanBe1000=DurationS==DurationS_1000?true:false;
+                        if (CanBe1001 && !CanBe1000)
+                            FrameRate_FromTags = float64_int64s(FrameRate_FromTags) / 1.001;
+                        if (CanBe1000 && !CanBe1001)
+                            FrameRate_FromTags = float64_int64s(FrameRate_FromTags) / 1.001001;
+                    }
 
                     // Duration from tags not reliable, checking TrackDefaultDuration
-                    if (!CanBe1000 && !CanBe1001)
+                    if (CanBe1000 == CanBe1001)
                     {
                         float64 Duration_Default=((float64)1000000000)/Temp->second.TrackDefaultDuration;
                         if (float64_int64s(Duration_Default) - Duration_Default*1.001000 > -0.000002


JeromeMartinez:
Please provide a sample file.

ndjamena:
Of the new tags or the frame rate thingy?

NUMBER_OF_BYTES_UNCOMPRESSED is being applied to subtitles compressed with zlib.  It's basically the stream size without the extra zlib compression.  I'm pretty sure it needs to be reformatted to look more like the proper stream size but I don't know how to do that.

SOURCE_ID is the ID of the track from the original blu ray.  ie the PID from the m2ts transport stream the track was taken from.


This is what's happening with the frame rate detection:
```
1 / 23.976         = 0.04170837504170837504170837504171
1 / (24 / 1.001) = 0.04170833333333333333333333333333

29820 * (1 / 23.976)        = 1243.7437437437437437437437437437
29821 * (1 / 23.976)        = 1243.7854521187854521187854521188
29820 * (1 / (24 / 1.001)) = 1243.7425
29821 * (1 / (24 / 1.001)) = 1243.7842083333333333333333333333

29820 * (1 / 23.976) -> Round=1243.744
29820 * (1 / (24 / 1.001)) -> Round=1243.743

1243.744 + (1 / 23.976) = 1243.785708375041708375041708375
1243.743 + (1 / (24 / 1.001)) = 1243.7847083333333333333333333333

1243.744 + (1 / 23.976) -> Round=1243.786
1243.743 + (1 / (24 / 1.001)) -> Round=1243.785

So there's the problem there.  Matroska doesn't have a frame rate and they only have a precision of one millisecond.  The way MKVMerge determines the final duration of a track is to take the last known time code and add the frame duration to it.  The last known time code is rounded, and the final duration is rounded as well.  In this case the last time code is round significantly up, and the final duration is rounded up as well.  In this case the rounding causes a disparity in the final duration of almost an four fifths of a millisecond.  Since the duration for 29821 frames at 23.976fps is rounded down, that difference is escalated to more than 1ms.  The difference between 23.976 and 24 / 1.001 is a millisecond every 16.666.... minutes.  So the actual difference at 20 minutes is much less than 2ms, rounded down it would in fact be 1ms.

With Matroska, on top of the rounding to milliseconds to deal with, a program also has to deal with the fact that the file may have been split at some point, which would mean all the roundings would occur in different places than expected, and further, some applications with access to the original frame rate may even take to applying the original duration instead of applying all this rounding.

Without more time codes to reference it is impossible to determine the difference between 23.976 and 24/1.001 in a file of that duration using nothing but the duration tag and frame count tag.
```

The code I've provided is just a simple hack, I'm not up to doing a more thorough job, I'd need to rewrite the entire thing.

JeromeMartinez:
> Of the new tags or the frame rate thingy?

Both (actually, any file you have with has a diffferent issue, so I can put them in my non regression tests)

ndjamena:
I don't mind providing them, it's the fact that it kills my internet connection for several hours to days that's the problem.

Since MKVMerge will remove it and not replace it, to keep the Source_ID I'd need to rip a file from an actual blu ray using MakeMKV, or I could just add the tag manually to any random file using MKVPropEdit.

I should be able to get the other tag by simply remuxing an MKV with MakeMKV.

If I encode 29821 blank frames using AVISynth and x264 I SHOULD be able to make a file that's detected as the wrong frame rate...

**sigh**

JeromeMartinez:
> it's the fact that it kills my internet connection for several hours to days that's the problem.

tags are at the beginning of the file, goal is no to test the coherency of the file (tags vs real data), so I don't need the complete file, providing e.g. the 1st MB is enough.

ndjamena:
... the pointer to the tags is at the beginning of the file ... the tags themselves are at the end ...

There was some one second m2ts files on the ST TNG discs I just ripped.  I wonder if they had their own playlists...

JeromeMartinez:
OK, so sorry but I need complete files.

ndjamena:
Nothing less than 10MB has subtitles in, but this is what the output looks like:

```
General
Unique ID                                : 218095247785994345798194904407971241132 (0xA4139DE97C5B27E08241DAFBDBA09CAC)
Complete name                            : F:\Make\Star Trek_ The Next Generation\Disc 23\00001.mkv
Format                                   : Matroska
Format version                           : Version 2
File size                                : 6.20 GiB
Duration                                 : 45mn 33s
Overall bit rate mode                    : Variable
Overall bit rate                         : 19.5 Mbps
Movie name                               : SegmentsMap:0,64
Encoded date                             : UTC 2015-12-21 06:11:56
Writing application                      : MakeMKV v1.9.8 win(x64-release)
Writing library                          : libmakemkv v1.9.8 (1.3.3/1.4.4) win(x64-release)

Video
Format                                   : AVC
Format/Info                              : Advanced Video Codec
Format profile                           : High@L4.1
Format settings, CABAC                   : Yes
Format settings, ReFrames                : 4 frames
Codec ID                                 : V_MPEG4/ISO/AVC
Duration                                 : 45mn 32s
Bit rate mode                            : Variable
Bit rate                                 : 1 791 Kbps
Maximum bit rate                         : 38.0 Mbps
Width                                    : 1 920 pixels
Height                                   : 1 080 pixels
Display aspect ratio                     : 16:9
Frame rate mode                          : Constant
Frame rate                               : 23.976 (24000/1001) fps
Color space                              : YUV
Chroma subsampling                       : 4:2:0
Scan type                                : Progressive
Bits/(Pixel*Frame)                       : 0.036
Stream size                              : 5.70 GiB (92%)
Language                                 : English
Default                                  : No
Forced                                   : No
Source ID                                : 001011

Audio
ID                                       : 2
Format                                   : FLAC
Format/Info                              : Free Lossless Audio Codec
Codec ID                                 : A_FLAC
Duration                                 : 45mn 33s
Bit rate mode                            : Variable
Bit rate                                 : 156 Kbps
Channel(s)                               : 8 channels
Channel positions                        : Front: L C R, Side: L R, Back: L R, LFE
Sampling rate                            : 48.0 KHz
Frame rate                               : 23.976 fps (2002 spf)
Stream size                              : 508 MiB (8%)
Title                                    : Surround 7.1
Writing library                          : MakeMKV v1.9.8 win(x64-release)
Language                                 : English
Default                                  : Yes
Forced                                   : No
Source ID                                : 001100

Text
ID                                       : 3
Format                                   : PGS
Muxing mode                              : zlib
Codec ID                                 : S_HDMV/PGS
Codec ID/Info                            : Picture based subtitle format used on BDs/HD-DVDs
Duration                                 : 44mn 25s
Bit rate                                 : 4 140 bps
Count of elements                        : 1436
Stream size                              : 5.86 MiB (0%)
Language                                 : English
Default                                  : No
Forced                                   : No
Stream Size (Uncompressed)               : 13796392
Source ID                                : 001200

Menu
00:00:00.000                             : en:Chapter 01
00:04:42.532                             : en:Chapter 02
00:10:57.990                             : en:Chapter 03
00:20:08.916                             : en:Chapter 04
00:26:00.725                             : en:Chapter 05
00:34:47.585                             : en:Chapter 06
00:44:42.262                             : en:Chapter 07
``
I haven't seen what Source_ID looks like from a DVD yet.  Stream Size (Uncompressed) should probably read  "13.80 MiB" (and possibly moved to just underneath Stream Size) but it would still need the field with the count of all the bytes as well, but hidden.

JeromeMartinez:
> Source_ID

I wonder the reason it is useful; shouldn't it be just discarded or hidden by default? and what is the format? decimal or hexadecimal or something else? the output is weird, because a bluray pid is 12-bit long (0x0000 to 0x1FFF)

> Stream Size (Uncompressed) should probably read "13.80 MiB" (and possibly moved to just underneath Stream Size)

Easy to add for me, will do. But I'll aslo set it as hidden by default for all formatings, this is not something the most important.

> but it would still need the field with the count of all the bytes as well, but hidden.

as all other fields, the raw value is always present.

***

Still need files (for all cases, including frame rate issue you have).

ndjamena:
This is a number from a DVD:

0180BD

This is a number from a BD:

001100

MediaInfo output for DVD

189 (0xBD)-128 (0x80)

and for BD:

4352 (0x1100)

I think the first Byte indicates the "Type" of ID being presented, ie 00 = BD ID, 01=DVD ID.

TrueHD and it's AC-3 core get the same ID, so do any duplicate re-encodes of a track, so the first byte has nothing to do with substreams or copies or the sort, yet it's marked as 01 in every stream from a DVD and 00 in every stream from a Blu Ray.

I believe the purpose is in fact to indicate two streams are identical.  TrueHD and its core is the most obvious case of identical streams in a single file, but MakeMKV allows a single stream to be re-encoded into an arbitrary number of different codecs simultaneously and copied as well to boot, all in a single pass.

The SourceID can serve two purposes.  1: it lets a user determine exactly where in the original blu ray the stream originated, and 2: it shows which streams contain identical content.  And it does it in a way that any piece of software would find easy to interpret.

MKVToolNix just implemented the output of the track identifiers of any supported containers in identification mode.  It outputs them all as 64bit numbers.  It's output of the same DVD track listed at the top of this post looks like this:

549755814077

From BD 80 it basically takes the BD, converts it to a 32 bit number, puts it in the lower 32 bits, then gets the 80 and puts that in the higher 32 bits.  It comes out as a uniform ID output for any container.  MediaInfo IDs are formatted differently for every container type, my code for translating MediaInfo to MKVMerge is to separate on the dash, then put the second part in the higher bits and the first in the lower.  I only have BD, MP4 and DVD to work with but so far it's turning out great, my little program can output all the info MediaInfo can give and the official MKVMerge TrackID to boot.  I wish I'd thought to ask for that last time I tried to merge the programs' output.  Matching MakeMKVs number to the original source shouldn't be hard either, and it does it with a single field with a single formatting.

Someone came to the MKVToolNix thread recently complaining that his files noticeably increased in
size after he'd remuxed it with MKVMerge.  He gave Mosu the MKVInfo output and was told he'd disabled zlib compression in the subtitles.  We had several people on VideoHelp looking at the DTS-MA output, ignoring the word "Unknown" and assuming the DTS-MA had a bitrate of "1509".  The word zlib means nothing to most people, I should think very few people even notice it's there...  but a number...  now that's different.

JeromeMartinez:
> I think the first Byte indicates the "Type" of ID being presented, ie 00 = BD ID, 01=DVD ID.

Very cryptic.

> MKVToolNix just implemented the output of the track identifiers of any supported containers in identification mode. 

Sample file please ;-).

>  The word zlib means nothing to most people, I should think very few people even notice it's there... but a number... now that's different.

It would be same if someone decodes an AVC stream and remux the raw video, the file size would be bigger.
This is not a reason for displaying the value of the "decoded" size.
I need to do choices about what to show by default, for the moment I am not convinced about showing it by default (but it is ok to add the field, and later I'll have an UI for selecting every line to display).

ndjamena:
And I just used my little program to find you this:
[MakeMKV Tags.zip](https://github.com/MediaArea/MediaInfoLib/files/68306/MakeMKV.Tags.zip)

From the idiots perspective, seeing the compression ratio of zlib is cool, and anyway

Bits/(Pixel*Frame)

What's the difference, really?

(I don't really care, it never occurred to me the zlib compression ratio even existed, much less would be nice to see, apparently neither did Mosu.  But now that I know it's a possibility it IS kind of cool.)

((If you want to economise...  forced flags on Video and Audio are pretty much useless, and the default flag on Video is rather questionable.))

JeromeMartinez:
> Bits/(Pixel*Frame)

this is a "end to end" information.
Issue with "size after uncompress" is that you have hte size of some "coded" (here, the PGS format). difficult to find the equivalent of a pixel with PGS (at least, for the moment I dont have width/height in pixels of the PGS stream, and there is no fixed frame rate)

>  forced flags on Video and Audio are pretty much useless, and the default flag on Video is rather questionable.

I definitely added it in the default display too quickly; I may think to remove them sooner or later.

JeromeMartinez:
> MakeMKV Tags.zip

Thanks. So it is OK for the SOURCE_ID and NUMBER_OF_BYTES_UNCOMPRESSED part of the patch.
but your Pull Request also contain something about the frame rate, so I can not merge the pull request until I have a sample file corresponding to the other part of the modified code.

ndjamena:
That's sad.  I just put an MKV in a zip and it compressed it from 745kb to 80kb.
[NTSC Film Killer.zip](https://github.com/MediaArea/MediaInfoLib/files/68385/NTSC.Film.Killer.zip)

Anyway, after a bit of fiddling (because both MakeMKV AND MKVMerge were setting the duration to the actual duration on first mux, and apparently MKVMerge was doing something to it to remind it what the duration should be - which may possibly have had something to do with the only block element in the file containing nothing but the very last frame, I had to go MKVMerge->MakeMKV->MKVMerge to get it working, removing the tags each time just in case).

Here is your sample.

ndjamena:
A pixel is a sample, a frame is a sample, an entire bloody movie is a sample.

zlib compresses the entire subtitle stream, there's a source size and a end size per sample, it's just that there's only a single sample to factor in.

JeromeMartinez:
I modified the output (https://github.com/MediaArea/MediaInfoLib/pull/72).
"SOURCE_ID" corresponding field is just below "ID" field. I also analyzed the tag for both DVD-Video and Blu-ray, please double check, and I also added the source name ("Blu-ray" or "DVD-Video") in the General part.
"NUMBER_OF_BYTES_UNCOMPRESSED" corresponding field is just below "StreamSize", with "conveion" to kB/GB/... but all fields are hidden by default (I still don't see the value added of this field if displayed by default) (I have a specific commit for setting it hidden by default, so if you want it visible by default in your on branch, just remove https://github.com/MediaArea/MediaInfoLib/commit/e06c95902ef30db99a7fd4f0245bd9ba9f83a715 in your branch).

ndjamena:
Hmmm.  I spent the time to write a little program to calculate how long it takes for 1000 and 1001, "rounded" and "actual" framerates to stop overlapping.

This is the results, (give or take some float80 precision errors):

Frame Rate | Last Overlapping Frame | Timecode of Frame
----------------|-----------------------------------|--------------
24|30949|00:21:30.832
30|40935|	00:22:45.865
60|78871|	00:21:55.832

And that doesn't take splitting into consideration.

There's also the point that the rounded and actual durations quite often differ by 1ms, the current code doesn't take that into account.  It's not a problem for the actual durations, but you could add a check to see wether the individual frame duration is rounded up or down and compensate accordingly.

I'd do it myself but my head is killing me at the moment.

PR #99, opened 2016-02-04
x #B977 Fix ChannelLayout for mov files with channel bitmap

pchapmanvc:
Bug 977 in the sourceforge issue tracker:
https://sourceforge.net/p/mediainfo/bugs/977/


JeromeMartinez:
I wonder how I could have put this "+1" in code, it makes no sense.
But unfortunately I have no test file for this piece of code, this is one the occasional pieces of code I implement with only the spec, without any file (at least I see no change with the files I have), would you mind to provide a sample file with such channel layout?

Issue #100, opened 2016-02-05
IMF XML file parsing broken when relative path used


pchapmanvc:
Currently if you pass an IMF XML file (ASSETMAP, CPL, PKL) to MediaInfoCLI as a relative path, then MediaInfo will not pick up the related XML files nor see the referenced sources.

e.g.
`mediainfo ASSETMAP.xml`
will not display any useful info about referenced files, but
`mediainfo /home/pete/ASSETMAP.xml`
will.

JeromeMartinez:
Thanks!
(I don't use relative paths enough, so I missed this issue)

PR #101, opened 2016-02-05
x Fix parsing of DCP MXF files when input file is not an absolute path

pchapmanvc:
GitHub issue #100

Previously, when an ASSETMAP/CPL/PKL file was given as a relative path,
MediaInfo would look for related assets in the wrong place because the
'directory' was empty yet it prepended a path separator anyway.

Where dirname is empty, we need to drop the path separator.

Signed-off-by: Peter Chapman <pchapman@vidcheck.com>


JeromeMartinez:
Thanks!
(I don't use relative paths enough, so I missed this issue)

PR #117, opened 2016-02-17
IMF: Ignore MarkerSequence whatever namespace prefix is used

pchapmanvc:
Hi again Jérôme,

Suggested change to remove assumption about what namespace prefix is used for MarkerSequence.


JeromeMartinez:
you're totally right, I apologize not having reused what you changed last week.

Issue #153, opened 2016-05-04
Relative paths to segment files for DASH MPD not handled correctly


jgreer:
For DASH MPD manifests, the source files are not always found correctly because the SourceDir can get added twice.

For example, if you have the following directory structure:
```
/tmp/files/dash/manifest.mpd
/tmp/files/dash/audio/en/init.mp4
/tmp/files/dash/audio/en/seg-1.mp4
/tmp/files/dash/audio/en/seg-2.mp4
/tmp/files/dash/video/720p/init.mp4
/tmp/files/dash/video/720p/seg-1.mp4
/tmp/files/dash/video/720p/seg-2.mp4
```

If you were to run `mediainfo /tmp/files/dash/manifest.mpd` then it works correctly, checking the segment files to see if they exist, then opening them to get whatever info it needs.

However, if you do `cd /tmp/files; mediainfo dash/manifest.mpd` then it will correctly check for the segment files to exist, but according to strace, it will try to open "dash/dash/video/720p/seg-1.mp4" (with "dash" doubled in there) when trying to read the segment files.

This seems to happen because File_DashMpd.cpp adds the filename to the sequence with its SourceDir, but then later in File__ReferenceFilesHelper.cpp when it's configuring the filenames, it assumes that filenames that don't start with a "/" need to have a full path added.  (I think.)

I have a pull request to avoid having File_DashMpd.cpp include the SourceDir when adding files to the sequence, but am not sure it is the correct solution.  In any case, I'll submit it for discussion.

PR #154, opened 2016-05-04
Don't include the SourceDir when adding a DASH segment file to the sequence.

jgreer:
The segment filename will get 'configured' later to include the SourceDir anyway, and that would duplicate the SourceDir in the path, so just include the base filename in the sequence instead.

### DIFF
--- a/Source/MediaInfo/Multiple/File_DashMpd.cpp
+++ b/Source/MediaInfo/Multiple/File_DashMpd.cpp
@@ -380,10 +380,12 @@ void template_generic::Decode()
                         Media_Name_Temp.insert(Time_Pos, Index);
 
                     Ztring File_Name;
+                    Ztring File_Name_With_Path;
                     if (!SourceDir.empty())
-                        File_Name+=SourceDir+PathSeparator;
+                        File_Name_With_Path+=SourceDir+PathSeparator;
                     File_Name+=BaseURL+Media_Name_Temp;
-                    if (!File::Exists(File_Name))
+                    File_Name_With_Path+=BaseURL+Media_Name_Temp;
+                    if (!File::Exists(File_Name_With_Path))
                         break;
                     Sequence->AddFileName(File_Name);
                     Index_Pos_Temp++;


jgreer:
The "configuration" I refer to happens here: https://github.com/jgreer/MediaInfoLib/blob/fix-relative-dash-paths/Source/MediaInfo/Multiple/File__ReferenceFilesHelper.cpp#L429-L465

jgreer:
Here's some debug output showing what happened before the fix:
```
$ mediainfo /tmp/files/dash/manifest.mpd
------------------------------------------
    stat("/tmp/files/dash/video/1/seg-94.m4f", {st_mode=S_IFREG|0664, st_size=181927, ...}) = 0
Adding filename = /tmp/files/dash/video/1/seg-94.m4f
...
Configuring file 94 with name /tmp/files/dash/video/1/seg-94.m4f
Configured as ............... /tmp/files/dash/video/1/seg-94.m4f
------------------------------------------


$ mediainfo dash/manifest.mpd
------------------------------------------
    stat("dash/video/1/seg-94.m4f", {st_mode=S_IFREG|0664, st_size=181927, ...}) = 0
Adding filename = dash/video/1/seg-94.m4f
...
Configuring file 94 with name dash/video/1/seg-94.m4f
Configured as ............... dash/dash/video/1/seg-94.m4f  <---- PROBLEM
------------------------------------------
```

And similarly the debug output after the fix:
```
$ mediainfo /tmp/files/dash/manifest.mpd
------------------------------------------
    stat("/tmp/files/dash/video/1/seg-94.m4f", {st_mode=S_IFREG|0664, st_size=181927, ...}) = 0
Adding filename = video/1/seg-94.m4f
...
Configuring file 94 with name video/1/seg-94.m4f
Configured as ............... /tmp/files/dash/video/1/seg-94.m4f
------------------------------------------


$ mediainfo dash/manifest.mpd
------------------------------------------
    stat("dash/video/1/seg-94.m4f", {st_mode=S_IFREG|0664, st_size=181927, ...}) = 0
Adding filename = video/1/seg-94.m4f
...
Configuring file 94 with name video/1/seg-94.m4f
Configured as ............... dash/video/1/seg-94.m4f
------------------------------------------
```

jgreer:
Potential fix for issue #153

JeromeMartinez:
I tested with the files I have, no regression so could be OK.
Is it possible to have a sample package demonstrating this issue?

jgreer:
Yep.  Here's a minimal set of files demonstrating the issue, along with output/logs with/without the patch.

Note: The sample files are a bit artificial, in that there should be 2 segment files, but I deleted the second file to make the issue obvious in the MediaInfo output, too.  This issue was more noticeable when the copy/paste startNumber bug was still there, because it would base the duration on how many segment files it could find -- so the duration would change based on what sort of path you used to reference the MPD file.

[mediainfo_path_bug_files.zip](https://github.com/MediaArea/MediaInfoLib/files/253048/mediainfo_path_bug_files.zip)

JeromeMartinez:
I see several issues, e.g. the "file" size (sum of all files size is also impacted), and a different behavior on Windows platform (no sum of all files size, duration is read from the init file).

I am OK for merging this PR if it fits your needs for the moment, but I may change the behavior of MediaInfo in the future when such case is met. I am not sure I want to trash the duration in the init file. Maybe I'll change the behavior and how the duration from the init file + display missing file names, I am not sure yet but I'll not handle the issue now;

jgreer:
Note: This isn't a critical issue for us, as we can always use full absolute paths in our systems.  (However it gets confusing when we forget to do that manually from the command line.)

By all means, choose a more appropriate fix - this was just a quick hacky attempt at fixing the immediate bug that it behaves differently on the same file when run from a different path, so figured you'd want to clear that up.  And I agree it would be nice to be able to differentiate between the manifest duration and content duration.  Thanks!

JeromeMartinez:
OK, let's go with this "hot fix" for the moment, I'll see later when I refactor the code how I can better handle it.

BTW, thanks for your time, eager to get more PR ;-).

PR #155, opened 2016-05-04
Don't try to print filename in XML output if input file doesn't exist.

jgreer:
If we do this, it segfaults, so don't do it.


jgreer:
Fixes https://sourceforge.net/p/mediainfo/bugs/991/

JeromeMartinez:
Argh, I forgot to test the vector size... :(.
Thanks

PR #159, opened 2016-05-09
Two patches for your review.

mikecancilla:
One mxf fix and one soft-telecine fix.  Thanks, Mike

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mxf.cpp
+++ b/Source/MediaInfo/Multiple/File_Mxf.cpp
@@ -5061,24 +5061,47 @@ bool File_Mxf::FileHeader_Begin()
 bool File_Mxf::Synchronize()
 {
     //Synchronizing
-    while (Buffer_Offset+4<=Buffer_Size && (Buffer[Buffer_Offset  ]!=0x06
-                                         || Buffer[Buffer_Offset+1]!=0x0E
-                                         || Buffer[Buffer_Offset+2]!=0x2B
-                                         || Buffer[Buffer_Offset+3]!=0x34))
+    //MXF magic number is 11 bytes: 06 0E 2B 34 02 05 01 01 0D 01 02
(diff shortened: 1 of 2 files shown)


JeromeMartinez:
I tested with my bunch of test files and I see the issues (I get some DTS-HD files with invalid MXF sync and I get some telecine streams detected)

There are several regressions with this PR.
MXF: File_Mxf::Synchronize() is used also for synchronization after a seek in the file, not only at the beginning. Additionally, MXF is also used in IMX (MPEG Video in MXF header in MOV, or Sony Acquisition Metadata in MXF in MXF)
Telecine: I don't remember exactly, maybe the idea was to avoid non telecine (when there is a change in the stream not telecine)

So I see that the PR shows several flaws in MediaInfo, but it is not acceptable as is. I'll test it more later this week and say how to have it accepted (issues fixed without creating new issues)

mikecancilla:
Thanks for your feedback.  Yes, I'm sure my fixes (like the MXF fix) possibly fail some other use cases, but for our use these two code changes cause no regressions and fix bugs.

Please keep me informed of the progress if possible.  Is there a mailing list?

Mike

JeromeMartinez:
>  Is there a mailing list?

No.
But for issues, you can open a [ticket](https://github.com/MediaArea/MediaInfoLib/issues).

> Please keep me informed of the progress if possible.

This is the goal of Pull Requests, we can debate here of the outcome ;-).

mikecancilla:
If you'd like me to open a ticket I can do that.  I'm not here to debate, just help.  :-)

JeromeMartinez:
> If you'd like me to open a ticket I can do that.

You asked for a mailing list. If it is only for this PR, no need to do something, we discuss here. Else (if it is not related to this PR), please open an issue ticket.

>  I'm not here to debate, just help. :-)

Debate in the meaning that I'll do a counterproposal and ask for feedback.
Don't hesitate to help more ;-).

PR #164, opened 2016-05-19
mediainfodll - python usage on linux

stan3:
Fix for usage on my machine (Ubuntu 16.04).

Without I get
```
Get with Stream=General and Parameter='FileSize'
Traceback (most recent call last):
  File "../Example/HowToUse_Dll.py", line 65, in <module>
    print(MI.Get(Stream.General, 0, "FileSize"))
  File "/srv/MediaInfoLib/Source/MediaInfoDLL/MediaInfoDLL.py", line 212, in Get
    return self.MediaInfoA_Get(self.Handle, StreamKind, StreamNumber, Parameter.encode("utf-8"), InfoKind, SearchKind).decode("utf_8", 'ignore')
ctypes.ArgumentError: argument 4: <class 'TypeError'>: wrong type
```
with python3  and python2 returns an empty string.

### DIFF
--- a/Source/MediaInfoDLL/MediaInfoDLL.py
+++ b/Source/MediaInfoDLL/MediaInfoDLL.py
@@ -180,7 +180,6 @@ class MediaInfo:
     MediaInfo_Count_Get.restype = c_size_t
 
     Handle = c_void_p(0)
-    MustUseAnsi = 0
 
     #Handling
     def __init__(self):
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
I am not comfortable with moving to the Unicode/w_char_p Python interface on Linux, i remember I got some problems with it (due to wchar_t is 4 bytes on Linux but Python uses 2-bytes? I don't remember well).
I prefer to keep the UTF-8 interface for the moment.

I see the issue with Python3 you try to resolve, but the problem is not resolved with the patch, just hidden.

the confusing variable should definitely go away.
I propose another fix for the other part, keeping Ansi/UTF-8 interface:
https://github.com/JeromeMartinez/MediaInfoLib/commits/Python

Does it resolve your issue?

stan3:
Thanks, yep, that makes it work for both python2 and 3.

PR #165, opened 2016-05-20
Python Get() API with Python3 on Linux/Mac.

JeromeMartinez:
Python binding: some errors during call of Get() API with Python3 on Linux/Mac.


JeromeMartinez:
From https://github.com/MediaArea/MediaInfoLib/pull/164

PR #166, opened 2016-05-23
add info about integer parameter

stan3:
Adding info about parameter per https://sourceforge.net/p/mediainfo/discussion/297610/thread/e4ca0ee0/


JeromeMartinez:
Thanks

PR #218, opened 2016-07-12
Add travis tests

g-maxime:


JeromeMartinez:
[Travis error](https://travis-ci.org/MediaArea/MediaInfoLib/jobs/144301988#L2230) is weird because there is #include "ZenLib/Conf.h" using namespace ZenLib; just before this line.

g-maxime:
But just out of the MediaInfoLib namespace...
The only place this variable is used is in the MediaInfoLib namespace at the line 1586 of the same file, any special reason to let it global ?

tribouille:
"using namespace ZenLib;" is not called in this case due to the compilation parameter... Those "defines" are not set : https://github.com/MediaArea/MediaInfoLib/blob/master/Source/MediaInfo/Video/File_Mpegv.cpp#L24
int64u is defined in #include "MediaInfo/Setup.h" which call #include "ZenLib/Conf.h".

@JeromeMartinez , do you prefer a global "using namespace ZenLib;" (top of the file) or just use ZenLib::int64u (better :) )?

tribouille:
The "matrix" is not correct.
The script trace.sh which tests the correctness of the trace won't work if you inactivate it (-DMEDIAINFO_TRACE_NO)

You will have to add this in a separate time in "script:"

JeromeMartinez:
> The only place this variable is used is in the MediaInfoLib namespace at the line 1586 of the same file, any special reason to let it global ?

No.
[Fixed](https://github.com/MediaArea/MediaInfoLib/pull/219)

PR #222, opened 2016-07-30
add missing buffer functions to Python interface

miguelgrinberg:
The buffer functions are missing in the Python DLL interface files.


GuillaumeRoques:
Thanks! We will take a look at your PR.

JeromeMartinez:
@miguelgrinberg would you mind to add a "by buffer" example like the [one in the .cpp](https://github.com/MediaArea/MediaInfoLib/blob/master/Source/Example/HowToUse_Dll.cpp#L97-L159) for [the .py example](https://github.com/MediaArea/MediaInfoLib/blob/master/Source/Example/HowToUse_Dll.py)?

JeromeMartinez:
Thanks.
Merging without example, we'll add the example later.

PR #250, opened 2016-10-17
Build shared lib by default

g-maxime:


JeromeMartinez:
I am OK for having shared lib the default for configure.ac, but AddThisToRoot_DLL_compile.sh should keep a single binary by default.
so  --enable-static --disable-shared as with CLI/GUI.

PR #251, opened 2016-10-18
Install libmediainfo-config and libmediainfo.pc

g-maxime:

### DIFF
--- a/Project/GNU/PKGBUILD
+++ b/Project/GNU/PKGBUILD
@@ -29,8 +29,5 @@ package() {
     install -dm 755 $pkgdir/usr/include/$i
     install -m 644 $srcdir/MediaInfoLib/Source/$i/*.h $pkgdir/usr/include/$i
   done
-  install -dm 755 $pkgdir/usr/lib/pkgconfig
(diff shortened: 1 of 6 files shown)


JeromeMartinez:
/debian and .spec should be updated at the same time in order to avoid duplicated code.

g-maxime:
Done

+ Removed sed commands (version already set by ./configure script)

JeromeMartinez:
> Removed sed commands (version already set by ./configure script)

Great (having it centralized is better), but I still see one "sed". Expected?

Project/GNU/PKGBUILD
```diff
@@ -29,8 +29,6 @@ package() {
     install -dm 755 $pkgdir/usr/include/$i
     install -m 644 $srcdir/MediaInfoLib/Source/$i/*.h $pkgdir/usr/include/$i
   done
-  install -dm 755 $pkgdir/usr/lib/pkgconfig
-  install -m 644 $srcdir/MediaInfoLib/Project/GNU/Library/libmediainfo.pc $pkgdir/usr/lib/pkgconfig
   sed -i -e 's|Version: |Version: '$pkgver'|g' $pkgdir/usr/lib/pkgconfig/libmediainfo.pc
```
g-maxime: Oups, ;)

And same for ZenLib:
```
Name: libzen
Version: 0.4.330.4.33
```

PR #252, opened 2016-10-19
Remove libmediainfo-config from devel rpm

g-maxime:


JeromeMartinez:
why?

g-maxime:
- Added by my previous PR
- Return build directory paths instead of install paths
- No present in the debian devel package

2016-10-19 13:50 GMT+02:00 Jérôme Martinez <notifications@github.com>:

JeromeMartinez:
We use mediainfo-config for config without install and .pc for config with install, ok.
(maybe something to change there, but let's keeping using this method for the moment)

PR #255, opened 2016-10-21
Small fix

pavel-pimenov:


JeromeMartinez:
Thanks!
(argh, I did not take the time to handle the crashes from the SF ticket... Definitely a problem that I can not handle all requests :( )

PR #261, opened 2016-10-25
BMP height is negative

pavel-pimenov:
https://sourceforge.net/p/mediainfo/feature-requests/485/

### DIFF
--- a/History_DLL.txt
+++ b/History_DLL.txt
@@ -20,6 +20,7 @@ Version 0.7.90
 + DXW: support of fake time code attribute
 x Teletext in MPEG-TS: CodecID, format (classic teletext or subtitle), video delay were sometimes not displayed
 x PDF: fixed crash with some files with full analysis set
+x #B485, BMP height is negative
 
 Version 0.7.89, 2016-09-30
 --------------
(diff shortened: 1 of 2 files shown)


JeromeMartinez:
As top down stuff does not everywhere (e.g. Firefox wrongly displays the image), can we add e.g.

    Fill(Stream_Image, 0, "Method", "Top down");

when the value is negative?

pavel-pimenov:
const char* Bmp_CompressionMethod(int32u CompressionMethod)
{
    switch(CompressionMethod)
    {
        case 0 : return "RGB";
        case 1 : return "RLE";
        case 2 : return "RLE";

Replace 
 case 1 : return "RLE8";
 case 1 : return "RLE4";

?

/* constants for the biCompression field */
#define BI_RGB        0L
#define BI_RLE8       1L
#define BI_RLE4       2L
#define BI_BITFIELDS  3L
#define BI_JPEG       4L
#define BI_PNG        5L

JeromeMartinez:
I imagined that 8 and 4 are bitdepth, but looks like it is wrong.
OK for changing to REL4/RLE8.

PR #269, opened 2016-10-26
Fix using namespace std

pavel-pimenov:


JeromeMartinez:
Build fails on Linux (it has no precompiled headers, need "use namespace std;" in the .cpp)

Source/MediaInfo/HashWrapper.h
```diff
@@ -49,7 +50,7 @@ class HashWrapper
         SHA512,
         HashFunction_Max,
     };
-    typedef bitset<HashFunction_Max> HashFunctions;
+    typedef std::bitset<HashFunction_Max> HashFunctions;
```
JeromeMartinez: why? using std::bitset; is set just above.

pavel-pimenov: CI - error https://travis-ci.org/MediaArea/MediaInfoLib/jobs/170764480
../../../Source/MediaInfo/Multiple/File_Mxf.h:734:5: error: ‘bitset’ does not name a type

     bitset<Stream_Max+1> StreamPos_StartAtZero; //information about the base of StreamPos (0 or 1, 1 is found in 1 file) TODO: per Essence code (last 4 bytes of the Essence header 0xTTXXTTXX)

     ^

JeromeMartinez:
I am not a big fan on putting "std::" prefix, from my point of view having xxx:xxx:xxx in C++ code makes code difficult to read.

I am OK for putting e.g. using std::bitset; but  I am reluctant to accept std:: stuff directly in classes (except when the class is public, this is not the case here)

Source/MediaInfo/HashWrapper.h
JeromeMartinez: my comment is about typedef std::bitset<HashFunction_Max> HashFunctions;

JeromeMartinez:
would you mind to remove std:: prefix and add using namespace std::bitset; when needed (here in HashWrapper.h and File_Mxf.h)?

pavel-pimenov:
Ok.

pavel-pimenov:
Remove prefix for all public std::bitset<> members?

Source\MediaInfo\File__Analyze.h
Source\MediaInfo\File__Analyze_MinimizeSize.h
Source\MediaInfo\Multiple\File__ReferenceFilesHelper_Sequence.h
Source\MediaInfo\Audio\File_Dts.h
Source\MediaInfo\Text\File_Eia608.h
Source\MediaInfo\Multiple\File_Mxf.h
Source\MediaInfo\Text\File_Teletext.h
Source\MediaInfo\MediaInfo_Config.h
Source\MediaInfo\MediaInfo_Internal.h
Source\MediaInfo\Reader\Reader_File.h

JeromeMartinez:
Ha ha. Good catch of incoherencies in my own coding rules.
I take it optionally, for me it is not a priority.

PR #300, opened 2016-11-09
V730 Not all members of a class are initialized (part-1)

pavel-pimenov:
Fix copy-paste
    int64u Duration;
    int64u UncompressedSize;
Fix V730:
V730 Not all members of a class are initialized inside the constructor. Consider inspecting: Duration, UncompressedSize. file_als.cpp 45
V730 Not all members of a class are initialized inside the constructor. Consider inspecting: Duration, UncompressedSize. file_ape.cpp 68
V730 Not all members of a class are initialized inside the constructor. Consider inspecting: Duration, UncompressedSize. file_rkau.cpp 41
V730 Not all members of a class are initialized inside the constructor. Consider inspecting: Duration, UncompressedSize. file_tta.cpp 41
V730 Not all members of a class are initialized inside the constructor. Consider inspecting: Channels, IsWB. file_amr.cpp 61
V730 Not all members of a class are initialized inside the constructor. Consider inspecting: BitDepth, Frequency, NumberOfChannelsMinusOne. file_pcm_vob.cpp 98
V730 Not all members of a class are initialized inside the constructor. Consider inspecting: VorbisHeader. file_tak.cpp 74
V730 Not all members of a class are initialized inside the constructor. Consider inspecting: StreamID. file__duplicate_mpegts.cpp 53
V730 Not all members of a class are initialized inside the constructor. Consider inspecting: ReferenceFilesHelper. file__referencefileshelper_common.h 30
V730 Not all members of a class are initialized inside the constructor. Consider inspecting: CountOfReferencesToParse, CountOfReferences_ForReadSize, StreamKind_Last, StreamPos_From, StreamPos_To. file__referencefileshelper.cpp 51


JeromeMartinez:
I am not a big fan of initializing all class members in the constructor, as lot of the members are used only if the parser is the correct one (e.g. if the file extension is the wrong one, I test all formats, and most of them will reject the file without using the members we are initializing with this PR).
But I understand that some QA tools complain.

is it worth it? I guess there is a very tiny speed impact in case the file extension is wrong (so all parsers are created).

pavel-pimenov:
Ok. revert PR.
p.s.
Create separate PR for?
int64u Duration;
int64u UncompressedSize;

imho init POD members very small overhead in many cases
Bonus:
* remove undefined behavior 
* random crash for T * members)
* less hours under debugging :)
```
class A
{
public:
    A(int x, int y) : x(x), y(y) { }
    int x, y;
    int* ptr;
};
int main(void)
{
    A* a = new A(3,3);
    std::cout << *a->ptr; // ups :)
}
```

JeromeMartinez:
with your example, there is also a crash if ptr is set to NULL ;-).

> Create separate PR for?
int64u Duration;
int64u UncompressedSize;

Let it as is for the moment, there is another issue with it (these fields are not part of the tag principle, this is only luck that all the corresponding audio classes use also the tag class).

pavel-pimenov:
Refactoring int64u Duration; and  int64u UncompressedSize
Add more C++;-)
https://github.com/MediaArea/MediaInfoLib/pull/303/files

PR #302, opened 2016-11-10
Remove unused int16u pid, int8u  stream_id;

pavel-pimenov:


JeromeMartinez:
argh, bad backport from me, I let some non open source stuff.
Can be removed, but please remove the "TODO" stuff else I'll have issue when I merge to the non open source version from a customer.
IgnoreGuardBandTest should also be removed.

pavel-pimenov:
Ok. 5 sec )

JeromeMartinez:
ok. I just pause the PR because I want to synchronize with another repo before accepting it.

PR #305, opened 2016-11-10
Modify channel and language index search logic

JeromeMartinez:
Old logic attempts to find a single index for both channel and language lookups.  Then for each referenced file, it splits the filename on "_", " ", "." and uses the calculated index to determine the channel and language metadata. This causes a segfault when a PTX session file references different length filenames and attempts to access an out of bounds index.

Instead, the fix attempts to find the channel and language index for each individual referenced file in the session.  Also added additional channels (Lsr, Rsr).


JeromeMartinez:
Initial PR from @res09ggm.

PR #323, opened 2016-11-29
Fix delete for void* (undefined behavior)

pavel-pimenov:
http://www.viva64.com/en/w/V772/

### DIFF
--- a/Source/MediaInfo/HashWrapper.cpp
+++ b/Source/MediaInfo/HashWrapper.cpp
@@ -62,8 +62,7 @@ static const char HashWrapper_Hex[16] = {'0','1','2','3','4','5','6','7','8','9'
 //---------------------------------------------------------------------------
 void HashWrapper::Init (const HashFunctions &Functions)
 {
-    m = new void*[HashFunction_Max];
-
+    memset(m,0,sizeof(m));
     #if MEDIAINFO_MD5
         if (Functions[MD5])
         {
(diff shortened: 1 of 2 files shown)


pavel-pimenov:
Replace ((void**)m)[xxх] -> m[xxx]  ?

JeromeMartinez:
I am not convinced by putting all pointers in the class.
Looking at the code, looks like the .h should be:

    void**  m;

Wouldn't be enough?

pavel-pimenov:
HashFunction_Max = 6
....
delete Hash; Hash=new HashWrapper(Config->File_Hash_Get().to_ulong());
...
double heap-allocations.
imho overhead?

JeromeMartinez:
> double heap-allocations.
imho overhead?

I don't understand, which overhead?

pavel-pimenov:
Small test:
https://github.com/pavel-pimenov/sandbox/commit/fe44eeb303d50ef9bf0673253e29e85f33a2b551

Q:\git\sandbox\console-test\memory>calc-ram.exe 1
RAM before = 2142208
RAM after= 7413760 Delta = 5271552 Type =  m[i] = new A1

Q:\git\sandbox\console-test\memory>calc-ram.exe 2
RAM before = 2109440
RAM after= 5783552 Delta = 3674112 Type =  m[i] = new A2

pavel-pimenov:
Close PR?

JeromeMartinez:
no close, I am just late on reviewing.

PR #332, opened 2016-12-16
Fix include

pavel-pimenov:


JeromeMartinez:
" 	Missing #include "MediaInfo/PreComp.h" and #pragma hdrstop"
I prefer not too touch to thirdparty code, and the precomp is not useful for these files.
please remove this commit.

JeromeMartinez:
there is an add then a remove. Better to just remove the commit.

pavel-pimenov:
git reset --hard 153a8bd692f7bdf0ebbb0f9172ff4676265e6635 ?

JeromeMartinez:
I don't know the command, I use an UI (TortoiseGit), I am not an expert, I use to save the last patch, "revert" to the first patch, then apply the last patch (so the middle patch is no more there).
Or store the 2 patches I want, use another branch an apply the 2 patches only.

pavel-pimenov:
ok

pavel-pimenov:
https://github.com/MediaArea/MediaInfoLib/pull/333

PR #336, opened 2016-12-18
Field handling

dericed:


JeromeMartinez:
I have some doubts about the second commit:
- the first one is the modification of an arbitrary decision (not in specs)
- the second one is the modification of names in the corresponding specifications

I understand that you try to avoid "/" because it is a separator for MediaTrace tests, but there is a conflict with the 2nd patch between "let's try to avoid slash" and "trying to keep names from specs".

If it is not blocking for the moment for the tests your are doing, I would be motivated for accepting only the 1st patch for the moment.

dericed:
2nd commit dropped

PR #344, opened 2016-12-26
Unicode[8] -> Unicode[3]

pavel-pimenov:
Max Pos = 2 (MB2WC(Handle, 2, OldParameter))


JeromeMartinez:
I guess it was a lazy method 15 years ago... :)

PR #347, opened 2016-12-30
Remove redundant std::map::operator[]

pavel-pimenov:


Source/MediaInfo/Audio/File_Aac_Main.cpp
```diff
@@ -490,20 +490,22 @@ void File_Aac::AudioSpecificConfig_OutOfBand (int64s sampling_frequency_, int8u
     {
         if (audioObjectType==(int8u)-1)
             return; //All data is not yet available
+        std::map<std::string, Ztring>::const_iterator i=Infos.find("Format_Settings_SBR");
         if (Infos.find("Format_Settings_SBR")!=Infos.end())
```
JeromeMartinez: should be if (i!=Infos.end()) now.

pavel-pimenov: Create new commit + squash?

JeromeMartinez: yes, please

PR #376, opened 2017-01-27
Valgrind: remove memory problems

tribouille:
Signed-off-by: Florent Tribouilloy <florent@mediaarea.net>


Source/MediaInfo/Video/File_Ffv1.cpp
```diff
@@ -474,6 +474,7 @@ File_Ffv1::File_Ffv1()
     ConfigurationRecordIsPresent=false;
     RC=NULL;
     version=0;
+    error_correction = false;
```
JeromeMartinez: error_correction is not a boolean, should be 0.

PR #398, opened 2017-02-21
EBUCore output is not correctly XML Encoded.

Zindar:
EBUCore output is not correctly XML Encoded.  This adds a few more XML_Encode to it.


JeromeMartinez:
Oops.
thanks.

PR #421, opened 2017-03-10
Added support using escape C string as output format.

kiyolee:
If someone wants to display details from multiple categories using just one command line, entering the ‘\n’ as format line separator is quite tricky and awkward.
On Windows, it is actually impossible to do that (at least in cmd.exe).
In bash, you may enter ‘\n’, but once you entered ‘\n’, the command line portion before ‘\n’ would become in-editable.
One alternative way is to complement mediainfo with printf command:
mediainfo --output="`printf 'General;%%CompleteName%%,\nAudio;%%BitRate%%'`" some.mp3
Still quite awkward however.
So I worked out the enhancement to support using Escape-C string as output format.
The above command line can become:
mediainfo --output="cstr:// General;%CompleteName%,\nAudio;%BitRate%" some.mp3
I choose the “cstr://” prefix just to minimize potential impact to existing interpretation of the output format string.
Hope this is useful.

### DIFF
--- a/Source/MediaInfo/MediaInfo_Config.cpp
+++ b/Source/MediaInfo/MediaInfo_Config.cpp
@@ -259,6 +259,138 @@ void MediaInfo_Config::Init()
 // Info
 //***************************************************************************
 
+static inline int _OctDigitValue(const String::value_type &ch)
+{
+    switch (ch)
+    {
+    case __T('0'): return 0;
+    case __T('1'): return 1;
+    case __T('2'): return 2;
+    case __T('3'): return 3;
+    case __T('4'): return 4;
+    case __T('5'): return 5;
+    case __T('6'): return 6;
+    case __T('7'): return 7;
+    }
+    return -1;
+}
+
+static inline int _HexDigitValue(const String::value_type &ch)
+{
+    switch (ch)
+    {
+    case __T('0'): return 0;
+    case __T('1'): return 1;
+    case __T('2'): return 2;
+    case __T('3'): return 3;
+    case __T('4'): return 4;
+    case __T('5'): return 5;
+    case __T('6'): return 6;
+    case __T('7'): return 7;
+    case __T('8'): return 8;
+    case __T('9'): return 9;
+    case __T('a'): case __T('A'): return 10;
+    case __T('b'): case __T('B'): return 11;
+    case __T('c'): case __T('C'): return 12;
+    case __T('d'): case __T('D'): return 13;
+    case __T('e'): case __T('E'): return 14;
+    case __T('f'): case __T('F'): return 15;
+    }
+    return -1;
+}
+
+static String _DecodeEscapeC(String::const_iterator first, String::const_iterator last)
+{
+    String decoded;
+    for (String::const_iterator it = first; it != last; ++it)
+    {
+        String::value_type ch = 0;
+        int inc = 0;
+        if (*it == __T('\\') && (it+1) != last)
+        {
+            switch (*(it+1))
+            {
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
I like the idea.
using a prefix is good because people may have used "\n" in their rules.
No better idea for the moment about how to deal with the "multiple categories using just one command line" request, I take this idea for the moment.

Source/MediaInfo/MediaInfo_Config.cpp
```diff
@@ -259,6 +259,142 @@ void MediaInfo_Config::Init()
...
+    case __T('b'): case __T('B'): return 11;
+    case __T('c'): case __T('C'): return 12;
+    case __T('d'): case __T('D'): return 13;
+    case __T('e'): case __T('E'): return 14;
+    case __T('f'): case __T('F'): return 15;
+    }
+    return -1;
+}
+
+String _DecodeEscapeC(String::const_iterator first, String::const_iterator last)
```
JeromeMartinez: please use "static"

Source/MediaInfo/MediaInfo_Config.cpp
```diff
@@ -295,6 +431,10 @@ Ztring MediaInfo_Config::Option (const String &Option, const String &Value_Raw)
         //Merge
         Value=FromFile;
     }
+    else if (Value_Raw.find(__T("cstr://"))==0)
```
JeromeMartinez: `Value_Raw` may be relatively big, doing a `find` may have performance impact with some corner cases, I am trying to avoid such impact in new code.
Please something e.g. [std::equal](http://stackoverflow.com/a/8596924/3882655).

Source/MediaInfo/MediaInfo_Config.cpp
kiyolee: Done.

Source/MediaInfo/MediaInfo_Config.cpp
kiyolee: Changed to substr() then compare.

JeromeMartinez:
using substr() is not what I would have done as there is a string copy, but this is a very tiny performance impact :), not important for the moment.
Please merge (or add) [the new commit](https://github.com/MediaArea/MediaInfoLib/pull/422/commits/dd0fb99d109f6262d2496728881c883929265391) in this PR, I would like to merge this PR as an independent PR.

kiyolee:
Hope the accidentally added files and then removed is not a problem.

JeromeMartinez:
Would you mind to merge the 3 commits in 1 so the files are not in the repo?

kiyolee:
Can you enlighten me how? Don't know that bit of git.

JeromeMartinez:
I am not a Git expert :(.
I use TortoiseGit [and combine commits in the UI](http://stackoverflow.com/a/21306306/3882655).

For Git command line, [Git squash](https://stackoverflow.com/questions/6934752/combining-multiple-commits-before-pushing-in-git).

Another option is to revert what you did and force push.

kiyolee:
Did git reset and force push. Hope this works.

JeromeMartinez:
2nd commit (the changes I requested) is missing ;-).

kiyolee:
Do that again. Better?

JeromeMartinez:
yes :).
Note that it will break the other PR (conflict because changes are not same about this feature) so you'll need to change it too after I merge this PR (after continuous-integration), sorry to ask for more work.

kiyolee:
Very well. Nice to give something back and learn something new.

PR #422, opened 2017-03-10
VS2017 support and other changes.

kiyolee:
1. Added VS2017 support. Just cloned from MSVC2015.
2. Keep various VS projects/solutions in sync with other (as much as possible). Fixed file format to what VS would generate.
3. zlib projects from official madler/zlib git-repo (and so your fork) do not build for x86. (__declspec related issue generating wrong symbol names that failed linking). Switched to use my kiyolee/zlib fork that works for both x86/x64.
4. Including my little enhancement for using Escape C string as output format. (Also as a seperate pull request)
Note: Need the corresponding pull request for ZenLib.


kiyolee:
AppVeyor build failed is expected as need to my zlib fork.

JeromeMartinez:
Moving to [kiyolee/zlib](https://github.com/kiyolee/zlib) would be great (I delegate VS projects handling :) ) but would break [our build scripts about zlib](https://github.com/MediaArea/MediaArea-Utils/tree/master/build_release/Diff) (we do a release with MT, XP and old 32-bit CPUs support so we apply a small .diff file).
I see that there is MT projects but I guess we would just do the same kind of patch as we already do with non MT projects)

@g-maxime any counter-argument to a move to a fork from [kiyolee/zlib](https://github.com/kiyolee/zlib) to [kiyolee/zlib](https://github.com/kiyolee/zlib)?

JeromeMartinez:
> Keep various VS projects/solutions in sync with other (as much as possible). Fixed file format to what VS would generate.

I appreciate, such stuff is not fun but useful.

JeromeMartinez:
Doing more checks about zlib part, I am less fan of switching to projects competing with zlib upstream, I would prefer a fix is in upstream. Anyway I take the PR as is and may changes some stuff in a near future.

JeromeMartinez:
@kiyolee do you plan to do same for MediaInfo repository?

PR #423, opened 2017-03-12
source::atsc_epg_block::event make noncopyable

pavel-pimenov:


JeromeMartinez:
why?
and it does not compile everywhere.

pavel-pimenov:
close PR

PR #426, opened 2017-03-14
Fix build with C++Builder 10.1 (Part 1)

pavel-pimenov:
1. [bcc32 Error] File_Mk.cpp(814): E2268 Call to undefined function 'fabs'
2. [bcc32 Error] algorithm(1430): E2093 'operator<' not implemented in type 'MediaInfoLib::File_Mpeg4::mdat_Pos_Type' for arguments of the same type


JeromeMartinez:
good to test with different compilers in order to catch bad code.
Here I guess I forgot a "static" before the operator, and the version with 1 argument is better.

PR #428, opened 2017-03-15
Refactoring MediaInfo_Config_MediaInfo

pavel-pimenov:


Source/MediaInfo/MediaInfo_Config_MediaInfo.cpp
```diff
@@ -1788,8 +1789,7 @@ Ztring MediaInfo_Config_MediaInfo::File_Duplicate_Set (const Ztring &Value_In)
     }
     if (!ToReturn.empty())
         ToReturn.erase(ToReturn.begin()); //Remove first ";"
-
-    CS.Leave();
+    }
```
JeromeMartinez: I am not a big fan of adding {}, but also not against it if you prefer it + fix compilation issues.

Source/MediaInfo/MediaInfo_Config_MediaInfo.cpp
```diff
@@ -1800,7 +1800,7 @@ Ztring MediaInfo_Config_MediaInfo::File_Duplicate_Get (size_t AlreadyRead_Pos)
     CriticalSectionLocker CSL(CS);
     if (AlreadyRead_Pos>=File__Duplicate_List.size())
         return Ztring(); //Nothing or not more than the last time
-    Ztring Temp=File__Duplicate_List[AlreadyRead_Pos];
+    const Ztring Temp=File__Duplicate_List[AlreadyRead_Pos];
```
JeromeMartinez: why not return File__Duplicate_List[AlreadyRead_Pos]; here?
(I missed something?)

Source/MediaInfo/MediaInfo_Config_MediaInfo.cpp
pavel-pimenov: CriticalSectionLocker: exception-safe code. (RAII)

JeromeMartinez: ok. don't forget the compilation error due to this change. (AppVeyor is complaining)

PR #430, opened 2017-03-16
Fix C++ Builder 10.1 error

pavel-pimenov:
E2354 Two operands must evaluate to the same type


JeromeMartinez:
I missed the point yesterday for the other patch, but would you mind to use spaces instead of tabs in your patches?

PR #501, opened 2017-05-14
Fix Mpega_Channels[]

pavel-pimenov:
V781 The value of the 'mode' variable is checked after it was used. Perhaps there is a mistake in program logic. Check lines: 1119, 1132. file_mpega.cpp 1119

default


JeromeMartinez:
Element_Begin1() is not followed by Element_End0().
I see this is a copy of a code you see few lines after this ones, and it is not good.
Please move this test at the beginning of audio_data_Layer3() (please delete the similar code you see, no more needed)

PR #511, opened 2017-05-25
ACELP typo correction

Sami32:


JeromeMartinez:
You need to change the [resulting cpp file](https://github.com/MediaArea/MediaInfoLib/blob/4468063f90bdf485a2d43b9fe66878dc37b9a15a/Source/MediaInfo/MediaInfo_Config_Automatic.cpp#L3759) too.

Sami32:
Yep...i did it too fast ;)
Thanks.

JeromeMartinez:
> Do the VIDEO_DELAY value can have decimal numbers ?

Yes (I think I don't do it for the moment, but it is a possibility if/when I want to have precision more than the millisecond)

Sami32:
Thank you :) It was for our parsing Regex string.

PR #517, opened 2017-06-01
FFV1: ConfigurationRecord trace in case of bad CRC, including CRC validity info

JeromeMartinez:


JeromeMartinez:
@dericed There are 2 cases:
- not enough place for CRC: you get "ConfigurationRecord size issue", and no "CRC-32" field, so "CRC-32" field presence test could be good.
- else "CRC-32" field as the same "OK"/"NOK" principle as FFV1 Slice CRC.

PR #540, opened 2017-06-16
FLV: support of FLV files with an ID3v2 header

JeromeMartinez:
Yes, it really exists...


JeromeMartinez:
[File from VLC issue tracker](https://trac.videolan.org/vlc/ticket/18397).

PR #542, opened 2017-06-18
 AAC: ADTS/ADIF duration in case of full parsing

JeromeMartinez:
AAC: ADTS/ADIF duration in case of full parsing and/or "risky bitrate estimation" option
and better support of non-48k frequencies

### DIFF
--- a/Source/MediaInfo/Audio/File_Aac.cpp
+++ b/Source/MediaInfo/Audio/File_Aac.cpp
@@ -151,13 +151,14 @@ void File_Aac::Streams_Update()
             switch(Mode)
             {
                 case Mode_ADTS    :
-                case Mode_LATM    : if (Config->File_RiskyBitRateEstimation_Get() && !adts_buffer_fullness_Is7FF)
+                case Mode_LATM    : if (Config->File_RiskyBitRateEstimation_Get() && !adts_buffer_fullness_Is7FF && (Config->ParseSpeed<1.0 || File_Offset+Buffer_Offset<File_Size))
                                     {
-                                        int64u BitRate=(Frequency_b/1024);
+                                        float64 BitRate=((float64)Frequency_b)/frame_length;
                                         BitRate*=aac_frame_length_Total*8;
                                         BitRate/=Frame_Count;
 
                                         Fill(Stream_Audio, 0, Audio_BitRate, BitRate, 10, true);
+                                        Fill(Stream_Audio, 0, Audio_Duration, (File_Size-Buffer_TotalBytes_FirstSynched)/BitRate*8*1000, 0, true);
                                     }
                                     break;
                 default           : ;
@@ -183,8 +184,10 @@ void File_Aac::Streams_Finish()
             Fill(Stream_Audio, 0, Audio_BitRate_Mode, "VBR", Unlimited, true, true);
             if (Config->ParseSpeed>=1.0)
             {
-                Fill(Stream_Audio, 0, Audio_BitRate_Minimum, ((float64)FrameSize_Min)/1024*48000*8, 0);
-                Fill(Stream_Audio, 0, Audio_BitRate_Maximum, ((float64)FrameSize_Max)/1024*48000*8, 0);
+                Fill(Stream_Audio, 0, Audio_BitRate_Minimum, ((float64)FrameSize_Min)/frame_length*Frequency_b*8, 0);
+                Fill(Stream_Audio, 0, Audio_BitRate_Maximum, ((float64)FrameSize_Max)/frame_length*Frequency_b*8, 0);
+                Fill(Stream_Audio, 0, Audio_SamplingCount, Frame_Count*frame_length);
+                Fill(Stream_Audio, 0, Audio_Duration, ((float64)Frame_Count)*frame_length/Frequency_b*1000, 0);
             }
         }
         else if (Config->ParseSpeed>=1.0)


Sami32:
As i accidentally discovered this PR, i enjoyed tested it :wink: 
It has worked on 4 files :-)
[twist_shout_SBR_v4.aac.txt](https://github.com/MediaArea/MediaInfoLib/files/1107227/twist_shout_SBR_v4.aac.txt)
[RoxioMediaManager_test4_AAC-LC_v4_Stereo_CBR_64kbps_44100Hz.aac.txt](https://github.com/MediaArea/MediaInfoLib/files/1107228/RoxioMediaManager_test4_AAC-LC_v4_Stereo_CBR_64kbps_44100Hz.aac.txt)
[AAC_8khz_Mono_5.aac.txt](https://github.com/MediaArea/MediaInfoLib/files/1107230/AAC_8khz_Mono_5.aac.txt)
[AAC_12khz_Mono_5.aac.txt](https://github.com/MediaArea/MediaInfoLib/files/1107229/AAC_12khz_Mono_5.aac.txt)

The problem is that the estimate duration is sligtly shorter than the real one, and the approximation should be superior to avoid playing issue with the renderers.
Because, as far as i tested, the renderers, over UPnP or DLNA, will play only until the duration sended if it is shorter than the media, or will stop when no audio is received, also if the duration is longer.

For a file of 30s, the duration is shorted of 1s, and for a file of 5'43 the estimate duration is shorter of 30s or 35s.

Though something weird seem to happen with the _RoxioMediaManager_test4_AAC-LC_v4_Stereo_CBR_64kbps_44100Hz.aac_ file.
The estimate duration is more than 10 times superior !

JeromeMartinez:
> the approximation should be superior to avoid playing issue with the renderers.

I disagree about what it should be in MediaInfo: I fully understand this criteria for your case, but MediaInfo has not this criteria.
And actually, as said explicitly, it is an estimation, you'll **always** have at least 1 file with a shorter duration than estimated, this is the risk of estimation.

MediaInfo policy is to display duration based on average bitrate of the first frames.
If you want "average bitrate of the first frames + 10%", this is you arbitrary choice and you do it after reading MediaInfo input (advantage: you change the delta you choose when you want :) ).

> For a file of 30s, the duration is shorted of 1s, and for a file of 5'43 the estimate duration is shorter of 30s or 35s.

This is not the issue: bitrate computing is based on the first 5 seconds, whatever is the duration of the file. If this 5 seconds are not similar to the other parts of the file, you get wrong info, that's all, whatever is the duration (10 seconds or 10 hours).

> The estimate duration is more than 10 times superior !

This is the reason this is said "risky" ;-).
This was wanted by a sponsor, I implement it (optional), with the risk due to estimation based on bitrate taken from only few seconds.
Do not use it if you don't accept this risk (in practice, sponsor's encoder uses "average bitrate" so it is fine for him, but in your case this is not usable)

***

Note that there are 2 cases handled in this PR: **risky** duration computed based on few first seconds bitrate, and full file parsing, I think the last one is more suitable for you but it is LONG parsing with long files (full parsing). Technically impossible to do differently, except some tweaking option (e.g. "OK,  seconds not good, maybe I check 10 seconds instead, of 1 minute, by option" (not yet done).

JeromeMartinez:
> LC_v4_Stereo_CBR_64kbps_44100Hz.aac file.
The estimate duration is more than 10 times superior !

At the start you get lot of silence (17 bytes per frame), then "normal" 180 bytes per frame.
bitrate is based on the 17-byte frames --> bam.
BTW, that means this is not CBR despite the file name ;-)
Need of more frames parsed for a better estimation (10 seconds, maybe)

Note that a feature for displaying bitrate by time slice is planned, but not ETA.

Sami32:
> MediaInfo policy is to display duration based on average bitrate of the first frames.
> If you want "average bitrate of the first frames + 10%", this is you arbitrary choice and you do it after reading MediaInfo input (advantage: you change the delta you choose when you want :) ).

Cool, i didn't thought about that :+1: 

I did understood the risky estimation, but do not evaluate how risky it could be LOL
> Need of more frames parsed for a better estimation (10 seconds, maybe)

That's seem a good idea to me.
> Note that a feature for displaying bitrate by time slice is planned, but not ETA.

I'll not throw you the stone, as i'm worst than you for "no ETA" :wink: 
I still need to try with the full parsing option.

I guess that it have something to do with `adts_buffer_fullness`, but i'm not sure to understand why some ADTS files can be parsed that way and some not ?
[id3v2_id3v1.aac.txt](https://github.com/MediaArea/MediaInfoLib/files/1112241/id3v2_id3v1.aac.txt)
[nocturne.aac.txt](https://github.com/MediaArea/MediaInfoLib/files/1112242/nocturne.aac.txt)
[test6_AAC-LC_v2_Stereo_VBR_128kbps_44100Hz.aac.txt](https://github.com/MediaArea/MediaInfoLib/files/1112243/test6_AAC-LC_v2_Stereo_VBR_128kbps_44100Hz.aac.txt)

JeromeMartinez:
> I guess that it have something to do with adts_buffer_fullness, but i'm not sure to understand why some ADTS files can be parsed that way and some not ?

It depends of the encoder: some do CBR for sure, encoding with lot of bytes the silence (they store the "noise" more precisely, this is the case for the sponsor of this feature, designed for broadcast with a constant satellite bandwidth) and some other keep the same quality during encoding (so silence is 17 byte long, this is the case with your files, designed for storage so constant quality is optimal).

This is normal, expected, AAC is by design VBR. Note that this is same with MP3, this is not new, difference is that lot of encoders put stats in the ID3v2 tag or LAME puts stats in the first frame with "hidden" bytes (and LAME was a lot used). Ask to your encoder providers to put ID3v2 tags ;-).

JeromeMartinez:
> > Need of more frames parsed for a better estimation (10 seconds, maybe)

> That's seem a good idea to me.

https://github.com/MediaArea/MediaInfoLib/issues please

Sami32:
Ah OK, so that could be a good point for us.
As most of my test files seem to have used Nero, SUPER, MediaConvert...as encoders, the users commercial AAC files have a greater probability to have ID3v2 tag included.

But _nocturne.aac_ seem to be CBR, and the bitrate is not estimate, so i've missed something.
**EDIT**: Aaah, i just used a russian tiny tool, and saw that it's not CBR file.
But MediaInfo don't say that it is VBR so i thought that it was a CBR file, by default deduction.

JeromeMartinez:
> But nocturne.aac seem to be CBR, and the bitrate is not estimate, so i've missed something.

`adts_buffer_fullness` is 0x7FF (highest value) so expected to be VBR, but it is definitely CBR (frames are between 371 and 372 byte long).
Currently I put "VBR" when `adts_buffer_fullness` is 0x7FF, I have doubts now...

> But MediaInfo don't say that it is VBR 

Hum, weird, is "VBR" with " --ParseSpeed=1", but nothing by default. A bug.
Looks like there is some review to do about coherency!
Closed PR is not a good place for speaking about different issues, please open an issue ticket per issue (e.g. "missing bitrate mode info"...), so I can better track issues and answer.

>  i thought that it was a CBR file, by default deduction.

When nothing is display on this line, that means that MediaInfo does not know, no other deduction is possible.

JeromeMartinez:
> the users commercial AAC files have a greater probability to have ID3v2 tag included.

`id3v2_id3v1.aac` has no bitrate / duration info, it does not count. Useful only if duration info is include.

Sami32:
Thank you for all these explanations :+1: 
OK, i'll. But my purpose was to test and report only, not to discover bugs ;-)

I tried with `ParseSpeed=1` and the bitrate look great, but i'm still struggling trying to understand which legacy fields `Legacy=false` disable ?

PR #570, opened 2017-07-05
Make template pkg-config file consistent between autotools and cmake

wiiaboo:
* Adding Libs_Static to CMake .pc, like in ZenLib
* zlib is a required dependency so it can be hardcoded into Requires
* -lpthread is pulled in from libzen, so it's not needed again

Should be the last modification required.


JeromeMartinez:
Thanks for the review and corresponding patches, appreciated.

wiiaboo:
Might want to add libcurl to autotools .pc file too, maybe?

JeromeMartinez:
All that make things better and coherent for everyone is appreciated!

PR #573, opened 2017-07-10
Update MediaInfo_Inform.cpp

Eld0r:
Added equal check to $if(...) statement in output generator.
Used online text editor; can somebody please compile and check this code?

And I just found an error in standard if else codeblock @779
I'm sorry it was late and I just wanted this feature..


JeromeMartinez:
why closed?

Eld0r:
I'm sorry it was late and I just wanted this feature so bad... There may be more principle problem for variables, too.
I'll first make sure my quickly written extension work. So it will not be embarrassing for me. :D

PR #576, opened 2017-07-18
pkg-config: fix build on debian/old RPM

g-maxime:
- zlib .pc isn't aviable on old systems
- pkg-config fail if libcurl .pc is missing


g-maxime:
On those systems libcurl is not dynamically loaded. and it a dependency for the dev packages, libmediainfo already require it for building.

libcurl is internal dependency of mediainfolib  and libcurl-dev package is needed only for static build (unfortunatly pkg-config fail if Require.private is not provided, even for dynamic build)

PR #587, opened 2017-08-07
Matroska: mapping of some additional tags to MediaInfo fields

JeromeMartinez:


JeromeMartinez:
@kieranjol may like this PR ;-).

kieranjol:
Hey! It's glorious! I was actually gonna ask earlier - it looks like it renames how the `TIMECODE` metadata tag appears in Matroska by using `Time code of first frame`, which is very nice too!

JeromeMartinez:
>  it looks like it renames how the TIMECODE metadata tag appears in Matroska by using Time code of first frame

Right, the idea in MediaInfo is to map different sources (here: a tag) to a common MediaInfo field not dependent of the source.
(and for HANDLER_NAME, looks like it comes from MOV, with same flaws, sometimes used for title and sometimes it is a generic name, I do as with MOV and hide it if it is a generic name)

kieranjol:
Can't wait to try this out the snapshot, thanks!

On 7 Aug 2017 16:35, "Jérôme Martinez" <notifications@github.com> wrote:

Merged #587 <https://github.com/MediaArea/MediaInfoLib/pull/587>.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/MediaArea/MediaInfoLib/pull/587#event-1195634386>, or mute
the thread

PR #590, opened 2017-08-07
EBUCore: WritingLibrary was missing if ReadByHuman option is set

JeromeMartinez:
Ugly but fast implementation before a complete refactoring of human readable fields management is implemented.


JeromeMartinez:
@g-maxime EBUCore and PBCore outputs should be same whatever is set the `ReadByHuman` option, please add some CI tests with XML.mxf and a file with video/audio in order to be sure we have this property always there.

PR #595, opened 2017-09-01
Remove include

pavel-pimenov:
default


JeromeMartinez:
Let's see if no compiler and version fails with that.

PR #597, opened 2017-09-02
Fix a weird 1.334 DAR due to a rounding issue

JeromeMartinez:
Note: @dericed I am aware this is a messy hot fix, and that I should spend time on storing Numerator and Denominator of PAR at some point instead of such workaround.


dericed:
"great patch Jerome"
mckayla-d7abd482-faa2-4ce6-8ee1-705a2101bc28

PR #598, opened 2017-09-03
FFV1: fix false positive about FFV1 slice_x error when slice_w>slice_h

JeromeMartinez:
@dericed which image will you find this time for this stupid bug? ;-)


dericed:
lol, am sure that @bturkus will be happy

PR #621, opened 2017-10-11
replace old FLAC website by the new one

baldurmen:
The main website for FLAC has been https://xiph.org/flac/ for a while now. This patch changes the old sourceforge subdomain to the new self-hosted one.


JeromeMartinez:
We definitively need to review links, thanks for doing it for FLAC.

PR #626, opened 2017-10-19
Old crappy XML output becomes OLDXML and MIXML becomes XML

JeromeMartinez:
And MIXML is promoted stable.


kieranjol:
This is a significant commit!

PR #630, opened 2017-10-23
cppcheck: use string::compare()

pavel-pimenov:


Source/MediaInfo/Multiple/File_Rm.cpp
```diff
@@ -369,7 +369,7 @@ void File_Rm::MDPR()
         CodecID_Fill(Ztring(mime_type.c_str()), Stream_Audio, StreamPos_Last, InfoCodecID_Format_Real);
         Fill(Stream_Audio, StreamPos_Last, Audio_Codec, "ralf");
     }
-    else if (mime_type.find("audio/")==0)
+    else if (mime_type.compare("audio/")==0)
```
JeromeMartinez: This was a lazy check which definitely needs to be changed, but `compare` returns 1 if `mime_type` is longer than "audio/", which is the case here.
Looks like the patch breaks the test.

PR #631, opened 2017-10-23
Fix V821

pavel-pimenov:
Decreased performance. The 'x' variable can be constructed in a lower level scope.
https://www.viva64.com/en/w/v821/print

### DIFF
--- a/Source/MediaInfo/Audio/File_Aac_GeneralAudio.cpp
+++ b/Source/MediaInfo/Audio/File_Aac_GeneralAudio.cpp
@@ -291,10 +291,10 @@ void File_Aac::program_config_element()
         if (!Infos["Format_Settings_SBR"].empty())
         {
             Infos["Format_Profile"]=__T("HE-AAC");
-            Ztring SamplingRate=Infos["SamplingRate"];
             Infos["SamplingRate"].From_Number((extension_sampling_frequency_index==(int8u)-1)?(Frequency_b*2):extension_sampling_frequency, 10);
             if (MediaInfoLib::Config.LegacyStreamDisplay_Get())
             {
+                const Ztring SamplingRate = Infos["SamplingRate"];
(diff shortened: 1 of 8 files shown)


Source/MediaInfo/Audio/File_Aac_GeneralAudio.cpp
```diff
@@ -306,13 +306,13 @@ void File_Aac::program_config_element()
...
         {
             Infos["Format_Profile"]=__T("HE-AACv2");
-            Ztring Channels=Infos["Channel(s)"];
-            Ztring ChannelPositions=Infos["ChannelPositions"];
-            Ztring SamplingRate=Infos["SamplingRate"];
             Infos["Channel(s)"]=__T("2");
             Infos["ChannelPositions"]=__T("Front: L R");
             if (MediaInfoLib::Config.LegacyStreamDisplay_Get())
             {
+                const Ztring Channels = Infos["Channel(s)"];
```
JeromeMartinez: `Channels` and `ChannelPositions` are modified in the meanwhile, so they don't behave same.
Fixed in https://github.com/MediaArea/MediaInfoLib/pull/703

Source/MediaInfo/Multiple/File_DvDif.cpp
```diff
@@ -1626,9 +1626,8 @@ void File_DvDif::video_recdate()
 
     Element_Name("video_recdate");
 
-    Ztring Date=recdate();
     if (Recorded_Date_Date.empty())
-        Recorded_Date_Date=Date;
+        Recorded_Date_Date= recdate();
```
JeromeMartinez: Issue is that recdate() is not called in all cases, and it parses a part of the stream so behavior is not exactly same.
But it is true that it is not the best method for parsing, I fixed the issue while changing the way it is implemented, in https://github.com/MediaArea/MediaInfoLib/pull/704.

PR #63, opened 2015-12-16
Removing warnings from autogen.sh.

jaeguly:

PR #633, opened 2017-10-23
Add FillInfosHEAACv2 fix #632

### LINKED ISSUE #632: Copy-paste Infos["Format_Settings_PS"].empty()
default

Create help function for consolidation and reduse code size?

pavel-pimenov:


JeromeMartinez:
It is not completely a copy/paste in the parts you delete, and the channels part is wrong with HEAACv2 streams ("2/2/2" instead of "2/2/1" meaning that an AAC-LC decoder will see only a mono stream), not usable as is.
There is definitely a refactoring to do, maybe more global, so either you are motivated for doing a bigger refactoring of all this mess (my code is messy, not yours :) ), or we wait a bit until I find some time for doing it.

JeromeMartinez:
The bug actually appeared in https://github.com/MediaArea/MediaInfoLib/pull/63, so I modify this PR in order to get the corresponding fix.
Modified version in https://github.com/MediaArea/MediaInfoLib/pull/703

PR #661, opened 2017-11-01
EBUCore: 1.8 Output uses now final version of XSD and final XSD location

JeromeMartinez:
https://github.com/MediaArea/MediaInfo/issues/197


kieranjol:
Awesome,we are going to be experimenting with the pb/ebucore outputs on
Friday.

JeromeMartinez:
> https://mediaarea.net/Services

New official version should be there in the meanwhile ;-).
Note that the JSON version has no XSD equivalent for the moment so no issue about that.

Issue #669, opened 2017-11-07
Matroska color range showing incorrect values


kieranjol:
Hi,

It seems that if a value < 2 is present in the Range element in a Matroska file the range is reported as Full.
I only just noticed this when making screenshots for my NTTW presentation..

Full range file -color_primaries jpeg
```
$ mediainfo --Details=1 jpeg.mkv |grep Range -A 4
01D5      Range - 2 (0x2) (4 bytes)
01D5       Header (3 bytes)
01D5        Name:                            5561 (0x15B9)
01D7        Size:                            1 (0x01)
01D8       Data:                             2 (0x02)
```
Broadcast range file -color_primaries mpeg
```
$ mediainfo --Details=1 mpeg.mkv |grep Range -A 4
01D5      Range - 1 (0x1) - Full (4 bytes)
01D5       Header (3 bytes)
01D5        Name:                            5561 (0x15B9)
01D7        Size:                            1 (0x01)
01D8       Data:                             1 (0x01)

```

Here's what the spec says:
`Clipping of the color ranges. (0: Unspecified, 1: Broadcast Range, 2: Full range (no clipping), 3: Defined by MatrixCoefficients/TransferCharacteristics)`

JeromeMartinez:
Argh, I stupidly mapped the Matroska value to the AVC/HEVC one (boolean 0 = Limited, 1 = Full).
Thanks for the bug discovery, need to fix that.

kieranjol:
Cool, I was going to try to fix it myself but it's well beyond my abilities.

JeromeMartinez:
Not so difficult, check https://github.com/MediaArea/MediaInfoLib/pull/670.
Note: I currently don't do any difference between 0 and 3, as I don't understand yet how `MatrixCoefficients/TransferCharacteristics` has an impact on range.

PR #670, opened 2017-11-07
Matroska: wrong color range info

JeromeMartinez:
Fix https://github.com/MediaArea/MediaInfoLib/issues/669

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mk.cpp
+++ b/Source/MediaInfo/Multiple/File_Mk.cpp
@@ -662,7 +662,16 @@ const char* Mpegv_colour_primaries(int8u colour_primaries);
 const char* Mpegv_transfer_characteristics(int8u transfer_characteristics);
 const char* Mpegv_matrix_coefficients(int8u matrix_coefficients);
 const char* Mpegv_matrix_coefficients_ColorSpace(int8u matrix_coefficients);
-extern const char* Avc_video_full_range[];
+
+const char* Mk_Video_Colour_Range(int8u range)
+{
+    switch (range)
+    {
+        case 1: return "Limited";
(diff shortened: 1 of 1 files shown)


Source/MediaInfo/Multiple/File_Mk.cpp
```diff
@@ -662,7 +662,16 @@ const char* Mpegv_colour_primaries(int8u colour_primaries);
 const char* Mpegv_transfer_characteristics(int8u transfer_characteristics);
 const char* Mpegv_matrix_coefficients(int8u matrix_coefficients);
 const char* Mpegv_matrix_coefficients_ColorSpace(int8u matrix_coefficients);
-extern const char* Avc_video_full_range[];
+
+const char* Mk_Video_Colour_Range(int8u range)
+{
+    switch (range)
+    {
+        case 1: return "Limited";
```
kieranjol: Is this terminology used across mediainfo,or is there a value to using the Matroska 0: Unspecified, 1: Broadcast Range, 2: Full range (no clipping):

kieranjol: *Matroska vocabulary I meant to say

JeromeMartinez: I use the first vocabulary I met, the MPEG one. I try to have the same vocabulary whatever is the source so I don't use Matroska termilogy (despite the fact it is more precise) and use the one I already have. So same policy in MediaConch whatever is the source.

kieranjol:
I understand,and it would actually be a lot easier for my own scripts too
to have the one mediainfo vocab.all LGTM anyhow,thx Jerome.

PR #674, opened 2017-11-10
Fixed undefined reference to File_Wtv

diegocr:
...which is caused by using `MEDIAINFO_WM_YES` instead of `MEDIAINFO_WTV_YES`


pavel-pimenov:
https://github.com/MediaArea/MediaInfoLib/blob/2892a7ab0685fd8a01288b7355400a7308958b4c/Source/MediaInfo/MediaInfo_File.cpp#L515-L517
?

diegocr:
Good catch :)

Fixed that one as well.

JeromeMartinez:
Oops... Thanks!
Don't hesitate to send patches for specific compilations options, it is useful.

PR #675, opened 2017-11-12
Fixed equality comparison that must be an assignment.

diegocr:
Summary says it all.

### DIFF
--- a/Source/MediaInfo/Tag/File_Id3v2.cpp
+++ b/Source/MediaInfo/Tag/File_Id3v2.cpp
@@ -980,7 +980,7 @@ void File_Id3v2::COMM()
     else if (Element_Values(0)==__T("MusicMatch_Mood"))
     {
         if (Retrieve(Stream_General, 0, General_Mood).empty())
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
Oops... Thanks for catching that!

diegocr:
Hi @JeromeMartinez, np! I'm building the library using Emscripten and it did thrown a few warnings which allowed me to catch :)

Btw, if you don't mind, could you please take a look at https://github.com/buzz/mediainfo.js/issues/14? perhaps you have some idea, which will be greatly appreciated since i'm really stuck on that issue with `Open_Buffer_Continue`

JeromeMartinez:
Planned but more complicated, longer to analyse, and I need some rest.
I did not care enough about a JS version of MediaInfo and I am very interested in having a stable version of it.

diegocr:
Ok, no problem, was just curious if you might be aware of some issue with `Open_Buffer_Continue` in 17.10 or it's just something of Emscripten, although a previous version did seem to work fine as available on that mediainfo.js repo - thanks in any case.

PR #686, opened 2017-11-19
Endianness and Sign report for PCM Matroska tracks

g-maxime:


JeromeMartinez:
Some issues with it:
- if BitDepth is before CodecID, crash
- it may have some differences between the File_Pcm instance creation and the cast part due to different test.

I reworked it in https://github.com/MediaArea/MediaInfoLib/pull/709.

PR #707, opened 2017-12-02
Fix build MediaInfoShellExt

pavel-pimenov:
MediaInfoShellExt_.cpp(5): fatal error C1083: Cannot open include file: 'MediaInfoDLL/MediaInfoDLL_Static.h': No such file or directory


JeromeMartinez:
I don't see the reason it would fix something if the directory name is `MediaInfoLib`.
But it is good to not depend of the directory name, so patch is relevant

PR #719, opened 2017-12-04
MOV: Support of external time code tracks (.qtc)

JeromeMartinez:
https://sourceforge.net/p/mediainfo/feature-requests/522/

+ support of file names >64 chars long or non ASCII for referenced files

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mpeg4_Elements.cpp
+++ b/Source/MediaInfo/Multiple/File_Mpeg4_Elements.cpp
@@ -3355,6 +3355,26 @@ void File_Mpeg4::moov_trak_mdia_minf_dinf_dref()
 }
 
 //---------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


Source/MediaInfo/Multiple/File_Mpeg4_Elements.cpp
```diff
@@ -3355,6 +3355,26 @@ void File_Mpeg4::moov_trak_mdia_minf_dinf_dref()
 }
 
 //---------------------------------------------------------------------------
+const size_t MacAlias_Size=16;
+const char* MacAlias[MacAlias_Size]
```
pavel-pimenov: static ...?

JeromeMartinez: updated

PR #724, opened 2017-12-07
MediaInfoLib Emscripten interface

g-maxime:

### DIFF
--- a/Project/GNU/Library/AddThisToRoot_DLL_compile.sh
+++ b/Project/GNU/Library/AddThisToRoot_DLL_compile.sh
@@ -22,16 +22,18 @@ Parallel_Make () {
     esac
     if [ "$numprocs" = "" ] || [ "$numprocs" = "0" ]; then
         numprocs=1
-    fi  
-    make -s -j$numprocs
+    fi
+   $Make -s -j$numprocs
 }
 
 ##################################################################
 # Init
 
 Home=`pwd`
+Make="make"
 ZenLib_Options=""
 MacOptions="--with-macosx-version-min=10.5"
+JsOptions="--host=le32-unknown-nacl"
 
 OS=$(uname -s)
 # expr isn’t available on mac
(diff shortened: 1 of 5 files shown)


JeromeMartinez:
a JS version was in our mind for a very long time and the [proof of concept](https://github.com/buzz/mediainfo.js) from @buzz confirmed that it is doable, thank you (note: API is not same, especially the seek part, and it is still work in progress, not considered stable until the official release).
@diegocr you may also be interested in this PR.

diegocr:
It's nice to have this as an option around, however like with the version from @buzz i'm seeing three problems here which would make it unusable on my company site:

1) The Emscripten bindings requires the use of `eval()` and/or `new Function()` which will make the library to fail on any site with a CSP excluding `unsafe-eval`, WebIDL must be used instead.

2) Making the lib consume a constant 1GB of memory is far from ideal, it will leave little room to do something else on any kind of advanced web app.

3) At first glance I think this specific version won't work with files > 2GB (the @buzz version had this fixed)

Just my 2 cents :)

JeromeMartinez:
1) It works for most usages so already a first step. We don't plan soon to use `unsafe-eval` on our side, we'll see if WebIDL is easy to use instead of `eval()` then we implement it if it is not too much work, else patch would be appreciated.

2) Right, but having 1 GB of memory usage is better than having nothing ;-), this is a first version. We plan to improve that (delivery date will depend on if we find sponsorship for that or not, there are definitely lot of places for memory usage reduction), but I understand (from other discussions / PRs) that you already did some work on that, how much memory usage reduction do you have?

3) At first glance I think this specific version won't work with files > 2GB

My tests on 100+ GB files say the opposite ;-). We used a different approach, converting `uint64_t` to `double` in the glue code instead of in the JS, so the limit is the JS limit (9007199254740991 bytes ~9 Po, I think we have some time before reaching the limit and JS may change before we reach it)

diegocr:
> It works for most usages so already a first step.

That is true indeed :)

> how much memory usage reduction do you have?

I didn't noticed any issue using 256MB, i even think that could be lower.

> My tests on 100+ GB files say the opposite ;-)

Ah, good to know, i wasn't really sure if the `double` trick would be sufficient here, I think to recall i did played with it on my build and it was giving me weird results, but it may have been due the glue code used.

JeromeMartinez:
> I didn't noticed any issue using 256MB, i even think that could be lower.

Tested with 128 MB and complex files, looks like it is OK.
I think I'll use this value, should be OK for most machines in 2017 and it may not worth to find a lower usable limit for the moment.

PR #738, opened 2017-12-13
Make File_VorbisCom.cpp dependant on MEDIAINFO_TAG_YES

diegocr:
This should just make an smaller build for those interested in having Ogg enabled without the bloat from the Tag parsers (I.e. `--disable-tag` used at configuring)


JeromeMartinez:
In theory, it would fail as VorbisCom is called from elsewhere even if MEDIAINFO_TAG_NO is defined.
In practice, MEDIAINFO_VORBISCOM_YES is not defined if MEDIAINFO_TAG_NO is defined, in Setup.h:

```
#if !defined(MEDIAINFO_TAG_NO) && !defined(MEDIAINFO_VORBISCOM_NO) && !defined(MEDIAINFO_VORBISCOM_YES)
    #define MEDIAINFO_VORBISCOM_YES
#endif
```

I think that the patch hurts, and does not minimize size.
Any scenario I missed?

diegocr:
First of all, let me point out that i have a somewhat huge patch with changes I made for our Emscripten's MediaInfo build, and i'm manually picking those changes that I think could be of help here/for you.

That said, you're correct, it's called from somewhere else, i did missed to submit some other changes on `File_Ogg_SubElement.cpp` related to this that should prevent the a compilation failure, and then the lib would compile fine without VorbisCom.

Here i do have I.e. `-DMEDIAINFO_OGG_YES=1 -DMEDIAINFO_VORBISCOM_NO=1`

Is that combination what was causing `File_VorbisCom.cpp` to be compiled even if we don't want it.

So, yeah this patch is intended to minimize the build size by not having VorbisCom support included.

JeromeMartinez:
> First of all, let me point out that i have a somewhat huge patch with changes I made for our Emscripten's MediaInfo build, and i'm manually picking those changes that I think could be of help here/for you.

Interesting. Very curious of the patches you have, don't hesitate to send as much as you can, I don't promise to accept all but I'll review all.
It would help us (thanks open source! we share work) and permit you to avoid incompatibilities when we update code.
Additionally, if it is relevant we could add a Travis CI job for testing specific compilation options (FYI we plan to add our own compilation options for our version of the JS build) so we are sure we break nothing when we have new PRs.

> i did missed to submit some other changes on File_Ogg_SubElement.cpp 

Looks like better, I'll test.
But I still think that the first commit is not relevant (it changes nothing), so should be removed.

> i'm not 100% sure if the #else part is quite correct, thanks.

It is fine, as `Open_Buffer_Continue` does something equivalent.

diegocr:
> Interesting. Very curious of the patches you have, don't hesitate to send as much as you can, I don't promise to accept all but I'll review all.

If you want I could send you by email the whole patch, and then we can decide which parts i should submit PRs for.

> But I still think that the first commit is not relevant (it changes nothing)

I may be missing something here, i think the file would be compiled in full if `MEDIAINFO_OGG_YES` is defined, regardless of the `MEDIAINFO_TAG_*` & `MEDIAINFO_VORBISCOM_*` defines ?

> It is fine, as Open_Buffer_Continue does something equivalent.

Cool, i was a little afraid of the validity of that part.

Source/MediaInfo/Tag/File_VorbisCom.cpp
```diff
@@ -17,7 +17,7 @@
 //---------------------------------------------------------------------------
 
 //---------------------------------------------------------------------------
-#if defined(MEDIAINFO_VORBISCOM_YES) || defined(MEDIAINFO_OGG_YES)
+#if defined(MEDIAINFO_TAG_YES) && (defined(MEDIAINFO_VORBISCOM_YES) || defined(MEDIAINFO_OGG_YES))
```
JeromeMartinez: Too much complex for no reason.
Replace by:

```
#if defined(MEDIAINFO_VORBISCOM_YES)
```

it is doable now thanks to your 2nd commit.

JeromeMartinez:
> If you want I could send you by email the whole patch, and then we can decide which parts i should submit PRs for.

Sure. do, please.

>  i think the file would be compiled in full if MEDIAINFO_OGG_YES is defined,

Right. See the review for how to fix that in a simpler way

Source/MediaInfo/Tag/File_VorbisCom.cpp
diegocr: That's even better, made it so - thanks for the hint!

PR #739, opened 2017-12-13
Fixed compilation issue at File__ReferenceFilesHelper.cpp

diegocr:
it does fail to compile if `MEDIAINFO_DVDIF_YES` is not defined.


JeromeMartinez:
thanks.

PR #740, opened 2017-12-13
Make Reader_File.cpp dependant on MEDIAINFO_FILE_YES

diegocr:
Just meant to create an smaller build.

### DIFF
--- a/Source/MediaInfo/Reader/Reader_File.cpp
+++ b/Source/MediaInfo/Reader/Reader_File.cpp
@@ -22,6 +22,10 @@
 //---------------------------------------------------------------------------
 
 //---------------------------------------------------------------------------
+#if defined(MEDIAINFO_FILE_YES)
+//---------------------------------------------------------------------------
+
+//---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
Ho... Right! good catch.

JeromeMartinez:
does not compile without `MEDIAINFO_READER_NO`.
https://github.com/MediaArea/MediaInfoLib/blob/master/Source/MediaInfo/MediaInfo_File.cpp#L766
https://github.com/MediaArea/MediaInfoLib/blob/master/Source/MediaInfo/MediaInfo_File.cpp#L1132
to be change to `MEDIAINFO_FILE_NO`.

JeromeMartinez:
After you send all your patches, I'll review the whole thing in order to see if all is coherent.
I prefer not to do it now in order not to break your patches. Let me know when all is finished on your side.

JeromeMartinez:
@diegocr such patch is included in https://github.com/MediaArea/MediaInfoLib/pull/749, as I needed to remove all code related to File API.

I try now to limit modifications in order not to conflict too much with pending patchs you have.

PR #744, opened 2017-12-15
Some colour description related updates

JeromeMartinez:
thanks to @kodabb for some hints.

### DIFF
--- a/Source/MediaInfo/Video/File_Mpegv.cpp
+++ b/Source/MediaInfo/Video/File_Mpegv.cpp
@@ -43,8 +43,8 @@ extern const char* Mpegv_colour_primaries(int8u colour_primaries)
         case  8 : return "Generic film";
         case  9 : return "BT.2020";                                     //Added in HEVC
         case 10 : return "XYZ";                                         //Added in HEVC 2014
-        case 11 : return "SMPTE RP 431-2";                              //Added in HEVC 2016
-        case 12 : return "SMPTE EG 432-1";                              //Added in HEVC 2016
+        case 11 : return "DCI P3";                                      //Added in HEVC 2016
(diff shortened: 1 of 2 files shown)


Source/MediaInfo/Video/File_Mpegv.cpp
```diff
@@ -63,11 +63,11 @@ extern const char* Mpegv_transfer_characteristics(int8u transfer_characteristics
         case  8 : return "Linear";
         case  9 : return "Logarithmic (100:1)";                         //Added in MPEG-4 Visual
         case 10 : return "Logarithmic (316.22777:1)";                   //Added in MPEG-4 Visual
-        case 11 : return "IEC 61966-2-4";                               //Added in AVC
+        case 11 : return "sRGB";                                        //Added in AVC
         case 12 : return "BT.1361 extended colour gamut system";        //Added in AVC
         case 13 : return "sYCC";                                        //Added in HEVC
-        case 14 : return "BT.2020 non-constant";                        //Added in HEVC
-        case 15 : return "BT.2020 constant";                            //Added in HEVC
+        case 14 : return "BT.2020 (10-bit)";                            //Added in HEVC, 10/12-bit difference is in ISO 23001-8
```
kodabb: it's also in hevc and avc: you could probably just drop the comments and just mention 23001 everywhere

JeromeMartinez: it is more for reference than for something else, I prefer "AVC" or "HEVC" because ITU versions are available for free and I saw them at this place, not the case of ISO 23001.

JeromeMartinez:
@kodabb following our discussion I added mapping of HDR10 display_primaries raw values to BT.709 and BT.2020, but it is theoretical as I don't have files (I hope I did not typo), would be great if you can test snapshots when available.

PR #749, opened 2017-12-17
Remove all code related to File API when MEDIAINFO_FILE_NO is set

JeromeMartinez:
Includes disabling MEDIAINFO_DUPLICATE too.


diegocr:
This is just awesome, thanks a ton for this!

The more smaller Emscripten builds we can make the better :)

JeromeMartinez:
> The more smaller Emscripten builds we can make the better :)

On our side, I think we'll stop for the moment, as our build is now ~ 1.1 MB gzipped.
A lot more could be done (lot of additional things could be changed, but it is bigger changes e.g. not using UTF-16 internally but change to UTF-8), but not our priority (no MediaInfo sponsor interested).

diegocr:
No worries, i may could help with that, i just need to get more familiar with the codebase, the patch i sent you is only a first step in that direction.

PR #760, opened 2017-12-20
Fix ReverseBits

pavel-pimenov:


JeromeMartinez:
Will not build if trace features are deactivated.
Please move it to before
#if !MEDIAINFO_TRACE
line.

PR #795, opened 2018-03-03
Hide/Show option

JeromeMartinez:
For hiding or showing some parameters.
Also add `FileNameExtension`

Example for showing "filename.ext" instead of "C:\\Downloads\\filename.ext"

```
mediainfo --HideParameter=General_CompleteName --ShowParameter=General_FileNameExtension
```


kieranjol:
Very cool!

PR #796, opened 2018-03-03
FFV1: remove sar_den test must be 0 if sar_num is 0

JeromeMartinez:
[Spec is currently not clear](https://github.com/FFmpeg/FFV1/issues/109) and FFmpeg creates such stream
(0/1 if unknown, e.g. with `./ffmpeg -f lavfi -i testsrc -vframes 1 -c:v ffv1 -level 3 -vf setsar=0/2 a.mkv`)

[FFV1-SLICE-sar_den test description](https://github.com/MediaArea/groundtruth/blob/master/matroska/54.md) needs to be updated.

[Reported by user](https://github.com/MediaArea/MediaConch/issues/210) and detected by me at the same period.

### DIFF
--- a/Source/MediaInfo/Video/File_Ffv1.cpp
+++ b/Source/MediaInfo/Video/File_Ffv1.cpp
@@ -1402,7 +1402,7 @@ bool File_Ffv1::SliceHeader(states &States)
         Param_Error("FFV1-SLICE-picture_structure:1");
     Get_RU (States, sample_aspect_ratio_num,                "sar_num");
     Get_RU (States, sample_aspect_ratio_den,                "sar_den");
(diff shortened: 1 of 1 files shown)


ablwr:
Detected by user in Archivemtica 1.7 too -- is there a way to mitigate this error prior to the next release?

JeromeMartinez:
No, it is hardcoded.
Note that in any case such file is a corner case and may be considered as problematic, as the DAR is unknown in that case, the file should be "fixed" even if it is valid technically speaking.

Note: [latest snapshot](https://mediaarea.net/download/snapshots/binary/mediaconch-gui) contains the fix, it could be used in Archivematica. Public release next week.

JeromeMartinez:
BTW, you need to clear the database, " --force" for each file with CLI (or drop the database), "Database" then "Delete all results from database" in GUI.

ablwr:
Thank you!

Issue #782, opened 2018-02-07
FFV1-SLICE-SliceContent fail


aeschweik:
Hi,

In running MediaConch on an FFV1 Matroska file I am getting a result I can't place. Running `mediaconch /Users/archivesx/Desktop/Annie/sample.mkv` gives the following results:

```
fail! /Users/archivesx/Desktop/Annie/sample.mkv
 -- FFV1-SLICE-SliceContent
 --    [fail:]
 --     [/Segment[1]/Cluster[1]/SimpleBlock[1]/Data[1]/Frame[1]/Slice[1]/SliceContent[1] at bytes offset of 745]
```

This does not happen with all FFV1 Matroska files I try, but I am not sure what would prompt this result.

Thanks!

JeromeMartinez:
I think it is a problem in our decoder (and this is not good for a checker!), as FFmpeg has no decoding issues.
thanks for the report, I'll dig in it soon.

privatezero:
@elderskatesman and the UW Libraries are reporting a similar issue with one FFV1/MKV out of a set they ran through mediaconch:

> FFV1-SLICE-sar_den Tests run: 4 | Results: ❌ Fail count: 4
Results: fail ❌ 
Name: sar_den
Offset: 1233
Value context: /Segment[1]/Cluster[1]/SimpleBlock[1]/Data[1]/Frame[1]/Slice[1]/SliceHeader[1]
Value: 1

JeromeMartinez:
[latest snapshot](https://mediaarea.net/download/snapshots/binary/mediaconch-gui) contains the fix. Public release next week, please review before the release if possible.
You need to clear the database, " --force" for each file with CLI (or drop the database), "Database" then "Delete all results from database" in GUI.

PR #798, opened 2018-03-06
FFV1: bug fixes and support of CRC in v0/v1

JeromeMartinez:
Fix for https://github.com/MediaArea/MediaInfoLib/issues/782
and consider also v0/v1 frames with CRC (`./ffmpeg -f lavfi -i mandelbrot -t 1 -c ffv1 -slicecrc 1 a.mkv`) as valid (todo: update spec with such case)

### DIFF
--- a/Source/MediaInfo/Video/File_Ffv1.cpp
+++ b/Source/MediaInfo/Video/File_Ffv1.cpp
@@ -830,6 +830,14 @@ void File_Ffv1::Read_Buffer_Continue()
         else
         #endif //MEDIAINFO_TRACE
             Skip_XX(Element_Size-Element_Offset,                "SliceContent");
+        if (version<=1 && Element_Offset+5==Element_Size)
+        {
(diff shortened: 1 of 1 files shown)


kieranjol:
Hi Jerome, what was the Golomb Rice parsing issue? I can't figure out from the C++...

JeromeMartinez:
I was "closing" (magic stuff `int8u s = 129; RC->get_rac(&s);`) the Range Coder part sometimes twice, sometimes 0 time in Golomb Rice mode due to code remaining from some early tests, it should be always once (note: [except with v0/v1 and Golomb Rice mode](https://github.com/JeromeMartinez/MediaInfoLib/blob/3cc05cba66ae338249bbfbdac63bd4203d299ab8/Source/MediaInfo/Video/File_Ffv1.cpp#L1251), I had to read FFmpeg code for understanding this exception, I need to document that in FFV1 spec), else the Golomb Rice part does not start at the right position.

PR #809, opened 2018-03-14
Ancillary data: fix infinite loop

JeromeMartinez:
Fix for https://sourceforge.net/p/mediainfo/bugs/1082/


ElSenorEd:
Thanks for the quick turnaround.

PR #815, opened 2018-03-15
V815 Decreased performance. (part2)

pavel-pimenov:
Consider replacing the expression 'string("")' with 'string()'. outputhelpers.cpp 98


JeromeMartinez:
Thanks for all the PRs, all merged now.

PR #829, opened 2018-03-31
File__analyze_streams.cpp - Change for 2.39:1 DisplayAspectRatio/Stri…

### LINKED ISSUE #612: DAR of 2.40:1 instead of 2.39:1
https://en.wikipedia.org/wiki/Aspect_ratio_(image)
> The motion picture industry convention assigns a value of 1.0 to the image’s height; an anamorphic frame (since 1970, 2.39:1) is often incorrectly described (rounded) as 2.40:1 or 2.40 ("two-four-oh").

> This aspect ratio of 2.39:1 was confirmed by the most recent revision from August 1993 (SMPTE 195-1993).[5]

http://www.arri.com/?eID=registration&file_uid=7737

Samples:
[audio-issue-sample-file.mkv.txt](https://github.com/MediaArea/MediaInfoLib/files/1308045/audio-issue-sample-file.mkv.txt)

kieranjol:
…ng #612 

This attempts to fix #612. I haven't tested this yet but I wanted to see if you thought this was sane.. It looks like two of us have a use case to have 2.39 display as intended anyhow..

### DIFF
--- a/Source/MediaInfo/File__Analyze_Streams.cpp
+++ b/Source/MediaInfo/File__Analyze_Streams.cpp
@@ -2507,7 +2507,8 @@ void File__Analyze::DisplayAspectRatio_Fill(const Ztring &Value, stream_t Stream
     else if (DAR>=(float)2.15 && DAR<(float)2.22)   DARS=__T("2.2:1");
     else if (DAR>=(float)2.23 && DAR<(float)2.30)   DARS=__T("2.25:1");
     else if (DAR>=(float)2.30 && DAR<(float)2.37)   DARS=__T("2.35:1");
-    else if (DAR>=(float)2.37 && DAR<(float)2.45)   DARS=__T("2.40:1");
+    else if (DAR>=(float)2.37 && DAR<(float)2.40)   DARS=__T("2.39:1");
+    else if (DAR>=(float)2.40 && DAR<(float)2.45)   DARS=__T("2.40:1");
     else                                            DARS.From_Number(DAR);
       DARS.FindAndReplace(__T("."), MediaInfoLib::Config.Language_Get(__T("  Config_Text_FloatSeparator")));
     if (MediaInfoLib::Config.Language_Get(__T("  Language_ISO639"))==__T("fr") &&   DARS.find(__T(":1"))==string::npos)


Source/MediaInfo/File__Analyze_Streams.cpp
```diff
@@ -2507,7 +2507,8 @@ void File__Analyze::DisplayAspectRatio_Fill(const Ztring &Value, stream_t Stream
     else if (DAR>=(float)2.15 && DAR<(float)2.22)   DARS=__T("2.2:1");
```
Sami32: Sorry, looking at your commit i just saw that 2.2:1 line above and get suspicious. i couldn't resist to post a remark.
It look to me that it should be `2.21:1`

> +The following fixed values are supported:
> +@table @option
> +@item 4/3
> +@item 16/9
> +@item 221/100
> +@end table
> +Any other value will result in square pixels being signalled instead
> +(see H.262 section 6.3.3 and table 6-3).

Taken from [there](https://git.ffmpeg.org/gitweb/ffmpeg.git/blobdiff/2bc9ba8d3c41f3a8e56484bd67b05040c7909a01..b78c30d7ec26af67c00ce2002709a189f6a87a7e:/doc/bitstream_filters.texi).

Source/MediaInfo/File__Analyze_Streams.cpp
```diff
@@ -2507,7 +2507,8 @@ void File__Analyze::DisplayAspectRatio_Fill(const Ztring &Value, stream_t Stream
     else if (DAR>=(float)2.15 && DAR<(float)2.22)   DARS=__T("2.2:1");
     else if (DAR>=(float)2.23 && DAR<(float)2.30)   DARS=__T("2.25:1");
     else if (DAR>=(float)2.30 && DAR<(float)2.37)   DARS=__T("2.35:1");
-    else if (DAR>=(float)2.37 && DAR<(float)2.45)   DARS=__T("2.40:1");
+    else if (DAR>=(float)2.37 && DAR<(float)2.40)   DARS=__T("2.39:1");
+    else if (DAR>=(float)2.40 && DAR<(float)2.45)   DARS=__T("2.40:1");
```
Sami32: Should we really keep that 2.40:1 value?

kieranjol: I dunno, it could break compatibility with some tools that are expecting that value? Though you could say the same for my commit, but I feel like this commit is closer to fixing a 'bug'. What is your reasoning for removing 2.40? That it is not a standardised ratio? I'd be in favour of keeping it ..

Source/MediaInfo/File__Analyze_Streams.cpp
kieranjol: Is this potentially outside of the scope of the pull request/issue which was dealing with the 2.39:1 issue?

Sami32: Hehe. As i said sorry, i couldn't resist when i discovered it ;)
You are right, i should have open an other issue. I'll try to remember to do that later.

Source/MediaInfo/File__Analyze_Streams.cpp
Sami32: Yes, because i'm always hesitant to leave not standardized stuff. Also if you are right that could break some program parsing code, as when we correct codec name or change the codec informations displayed, i consider that the tools, programs using MediaInfo have to adapt.

That said, in that case that doesn't hurt AFAIK and it is only one line of code so...

Source/MediaInfo/File__Analyze_Streams.cpp
kieranjol: :)

JeromeMartinez:
>  I wanted to see if you thought this was sane.. 

It is sane, sure, and IMO it is bug fix so "breaking" (it should break nothing as I expect people to use the raw vvalue) old stuff is fine.

I let you both debate while I enjoy some daya AFK :).

JeromeMartinez:
>  I wanted to see if you thought this was sane.. 

It is sane, sure, and IMO it is bug fix so "breaking" (it should break nothing as I expect people to use the raw vvalue) old stuff is fine.

I let you both debate while I enjoy some days AFK :).

kieranjol:
Ha! Enjoy the AFK - I think @Sami32 and I both agree on the content of this particular pull request anyhow, perhaps the other comments can be resolved in the other PR by @Sami32 ?

PR #842, opened 2018-04-30
Fix Identical condition

pavel-pimenov:
'moov_trak_mdia_minf_stbl_stsd_Pos>1' second condition is always false


JeromeMartinez:
thanks.

PR #844, opened 2018-04-30
Remove unused static function

pavel-pimenov:


JeromeMartinez:
I am not sure I would like to remove them, maybe it is better that I use them, not normal that the corresponding parsing code is not there.

PR #845, opened 2018-04-30
PBCore 2.1 updates

dericed:


Source/MediaInfo/Export/Export_PBCore2.cpp
```diff
@@ -179,8 +179,24 @@ Ztring ToReturn;
...
+        else if (!MI.Get(Stream_Video, StreamPos, Video_Delay_String4).empty())
+            Node_EssenceTrack->Add_Child_IfNotEmpty(MI, Stream_Video, StreamPos, Video_Delay_Original_String4, "essenceTrackTimeStart", "annotation", std::string("from container"));
+        else if (!MI.Get(Stream_Video, StreamPos, Video_Delay_String3).empty())
+            Node_EssenceTrack->Add_Child_IfNotEmpty(MI, Stream_Video, StreamPos, Video_Delay_Original_String3, "essenceTrackTimeStart", "annotation", std::string("from container"));
+    }
+
     //essenceTrackDuration
-    Node_EssenceTrack->Add_Child_IfNotEmpty(MI, StreamKind, StreamPos, "Duration_String3", "essenceTrackDuration");
+    if (!MI.Get(StreamKind, StreamPos, __T("Duration_String4")).empty())
+        Node_EssenceTrack->Add_Child_IfNotEmpty(MI, StreamKind, StreamPos, "Duration_String4", "essenceTrackDuration");
```
kieranjol: +1

Source/MediaInfo/Export/Export_PBCore2.cpp
```diff
@@ -270,11 +286,18 @@ Ztring Export_PBCore2::Transform(MediaInfo_Internal &MI, version Version)
 
     Node_Main.Add_Child("instantiationIdentifier", instantiationIdentifier, "source", "File Name");
 
-    // need to figure out how to get to non-internally-declared-values
-    //if (!MI.Get(Stream_General, 0, General_Media/UUID).empty())
-    //{
-    //    ToReturn+=__T("\t<instantiationIdentifier source=\"Media UUID\">")+MI.Get(Stream_General, 0, General_Media/UUID)+__T("</instantiationIdentifier>\n");
-    //}
+    // call UniqueID as SegmentUID for Matroska, else UniqueID
```
kieranjol: Haven't tested but great idea

Source/MediaInfo/Export/Export_PBCore2.cpp
```diff
@@ -356,13 +356,15 @@ Ztring Export_PBCore2::Transform(MediaInfo_Internal &MI, version Version)
...
     Node_Main.Add_Child("instantiationDigital", Format);
-    
+
     //formatStandard
     Ztring formatStandard=MI.Get(Stream_General, 0, General_Format);
     if (!MI.Get(Stream_General, 0, General_Format_Commercial_IfAny).empty())
         formatStandard+=__T(" (")+MI.Get(Stream_General, 0, General_Format_Commercial_IfAny)+__T(")");
-    Node_Main.Add_Child("instantiationStandard", formatStandard);
-    
+    Node* Child=Node_Main.Add_Child("instantiationStandard", formatStandard);
```
kieranjol: Again, i haven't tested this but adding these annotations makes sense as it somewhat mirrors the annotations in essenceTrackEncoding

JeromeMartinez:
```
		<essenceTrackIdentifier source="ID (Mediainfo)">1</essenceTrackIdentifier>
		<essenceTrackIdentifier source="UniqueID (Mediainfo)">1</essenceTrackIdentifier>
		<essenceTrackIdentifier source="StreamKindID (Mediainfo)">0</essenceTrackIdentifier>
		<essenceTrackIdentifier source="StreamOrder (Mediainfo)">0</essenceTrackIdentifier>
```

I don't like the idea to have " (MediaInfo)", ID and UniqueID are not "created" by MediaInfo, they are from the file. "ID" and "UniqueID" without " (MediaInfo)"  would be better IMO. for StreamKindID and StreamOrder it is guessed from the file, but less obvious.

PS: "MediaInfo", not "Mediainfo" ;-).

JeromeMartinez:
```
<instantiationDataRate unitsOfMeasure="bits/second">468348</instantiationDataRate>
```

https://en.wikipedia.org/wiki/Data-rate_units indicates that unitsOfMeasure should be "bit/s"

```
<instantiationFileSize unitsOfMeasure="bytes">58602</instantiationFileSize>
```

I think that unitsOfMeasure is singular (e.g. from https://en.wikipedia.org/wiki/Bit_rate it is said "1 kbit/s = 1,000 bit/s", not "bits/s"), and for an attribute it is even more obvious IMO --> "byte" without "s".

PR #848, opened 2018-05-05
Pbcore 2.1 updates

dericed:


dericed:
responds to comments at https://github.com/MediaArea/MediaInfoLib/pull/845#issuecomment-385637364 and https://github.com/MediaArea/MediaInfoLib/pull/845#issuecomment-385638137

PR #862, opened 2018-06-01
Fix C++Builder 10.1

pavel-pimenov:
[bcc32 Error] MediaInfo_Config.h(231): E2015 Ambiguity between 'MediaInfoLib::Char' and 'ZenLib::Char'
 [bcc32 Error] Export_Mpeg7.cpp(1229): E2354 Two operands must evaluate to the same type


JeromeMartinez:
thanks!

PR #869, opened 2018-06-11
fla

Sami32:
Complement this [MediaInfo PR](https://github.com/MediaArea/MediaInfo/pull/252).

### DIFF
--- a/Source/Resource/Text/DataBase/Format.csv
+++ b/Source/Resource/Text/DataBase/Format.csv
@@ -23,7 +23,7 @@ ISM;Internet Streaming Media;;M;Ism;;ism;;;
 IVF;;;M;Ivf;;ivf;;;
 LXF;;;M;Lxf;;lxf;video/lxf;;
 Matroska;;;M;Mk;;mkv mk3d mka mks;;https://matroska.org/downloads/windows.html;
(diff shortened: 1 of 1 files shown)


Sami32:
I shouldn't have added VRO as a specific format, perhaps as MPEG-PS.
I've done that from memory and it seem that i cannot rely on it ;)
http://www.pixelbeat.org/docs/DVD-VR/
https://en.wikipedia.org/wiki/DVD-VR

Sami32:
Shouldn't **DTS** be "Dedicated To Sound" instead of "Digital theater Sound"?
https://dts.com/
https://en.wikipedia.org/wiki/DTS_(sound_system)
https://ccm.net/contents/46-dts-dedicated-to-sound

JeromeMartinez:
> Shouldn't DTS be Dedicated To Sound instead of "Digital Theater Sound"?

Marketing change ("formerly known as Digital Theater Systems, Inc."), I'll change it at some point.

Sami32:
I asked because i was not sure you wanted me to add this change into this PR.
So, if you agree i could fix it here.

JeromeMartinez:
of for changing DTS thing.

Sami32:
Ok, so can i leave the DTS Wikipedia link or should i remove it?

JeromeMartinez:
Superseeded by https://github.com/MediaArea/MediaInfoLib/pull/869.

PR #870, opened 2018-06-11
SPDIF

Sami32:
Complement to this [MediaInfo PR](https://github.com/MediaArea/MediaInfo/pull/253).

### DIFF
--- a/Source/Resource/Text/DataBase/Format.csv
+++ b/Source/Resource/Text/DataBase/Format.csv
@@ -98,6 +98,7 @@ RKAU;RK Audio;;A;Rkau;;rka;;http://www.msoftware.co.nz;
 Scream Tracker 3;;;A;S3m;;s3m;;;
 Shorten;;;A;;;shn;;http://etree.org/shnutils/shorten/;Lossless
 SLS;;;A;;MPEG-4 Scalable Lossless Coding;sls;;http://www.chiariglione.org/mpeg/technologies/mp04-sls/index.htm;Lossless
+S/PDIF;;;A;Spdif;Sony/Philips Digital Interface;spdif;;https://en.wikipedia.org/wiki/S/PDIF;
 Speex;;;A;;;spx;audio/speex;http://www.speex.org/;Lossy
 Opus;;;A;;;;audio/opus;http://opus-codec.org/;Lossy
 TAK;;;A;;;tak;;http://thbeck.de/Tak/Tak.html;Lossless


JeromeMartinez:
Argh, I should not have merged it.
It will never be used, as MediaInfo detects the technical name ("SMPTE ST 337" a.k.a. AES3).
I'll check how I change that, having the link on S/DPIF page is not so bad.

Sami32:
Sorry, i should have said that i wasn't sure about that one.

Sami32:
I feel uncorfortable how to handle this format.

IEC 60958-1:2014 p.25
```
The IEC 60958 series is based on two different industry standards: the AES/EBU digital audio 
interface  standard  (AES3  and  EBU  Tech.  3250-E)  and  the  digital  interface  specification  by 
Sony and Philips (Sony-Philips Digital Interface Format (SPDIF)) introduced with the Compact 
Disc Digital Audio system.
Unfortunately,  significant  differences  between  the  two  standards  exist,  which  can  contribute  in 
part  to  the  different  application  areas:  professional  and  consumer.  The  differences  have 
contributed to many misunderstandings about the use and compatibility of the standards.
```

Do any difference exist between [SMPTE ST 337](https://ieeexplore.ieee.org/document/7291671/) and SMPTE 337M? I admit having no clue on that.

Sami32:
Hum...should really SMPTE ST 337 be the correct format to be displayed?
[DTSHD.spdif.txt](https://github.com/MediaArea/MediaInfoLib/files/2091025/DTSHD.spdif.txt)

I'm wondering if SPDIF shouldn't have been more correct?
As we could have displayed "IEC 61397" instead in the sample above?

That's said, as i'm not a standard expert as you, i prefer leave it to you ;)

JeromeMartinez:
> I'm wondering if SPDIF shouldn't have been more correct?

Long story short:
- Your file is **not** SPDIF dump. It is missing the preamble and VUCP flags (see https://en.wikipedia.org/wiki/AES3#Relation_to_S/PDIF and after this paragraph)
- Your file is a dump of the audio samples part
- the audio samples part contains usually raw data (any data, not detectable) for stereo, and for more than 2 channels they use SMPTE ST 337 for transporting non PCM content. Reason it is flagged SMPTE ST 337.

I have some AES3/SPDIF dumps (with preamble and VUCP flags), they are different (and not supported by MediaInfo currently).
o you'll never hit this line (from the PR) for the moment (and it does not hurt).

Sami32:
Ok, my bad then. I've produced it with _FFmpeg_, and my poor knowledge get me fooled.
Sorry for that wrong SPDIF line of code having been merged 😞 
As SMPTE ST 337 is a standard it seem strange to me to get it displayed as a format, but i'm not the expert here so it was more a wondering.

The good point for me is that now i should be aware of the true SPDIF nature :)

JeromeMartinez:
> As SMPTE ST 337 is a standard it seemed strange to me to get it displayed as a format,

VC-1 is a standard and a format name
H.265 is a standard and a format name
HEVC is a standard and a format name
Actually, what is not a format name when not a standard?
"Format" is synonym of standard name, just that SMPTE does not put fancy names and use numbers (as ITU does for "H.265", BTW, each letter and digit has a significance, like... SMPTE method).
In short, "format" field needs content, and I try to find the best suitable name (sometimes it is not easy, some competing names e.g. "High Efficiency Video Coding", "HEVC", "H.265", "MPEG-H Part 2", "ISO/IEC 23008-2" are different names for the same thing)

Sami32:
Hehe, ok, i'll try to keep it in mind.
That one one was just unknown from me and very long ;)

Issue #897, opened 2018-07-04
ATRAC9 format not identified correctly


Sami32:
Seing that _FFmpeg_ have just added the ATRAC9 decoder i tested MediaInfo and discovered that the Sony ATRAC format was not handled correctly by MediaInfo:

```
General
Complete name                            : F:\_DEVELOPPEMENT_TOOLS\!MEDIA_TEST_FILES\AUDIO\snd0.at9
Format                                   : Wave
File size                                : 1.62 MiB
Duration                                 : 1 min 10 s
Overall bit rate                         : 192 kb/s
FileExtension_Invalid                    : act wav

Audio
Format                                   : BA368D4D-E147-D242-88FC-61654F8C836C
Codec ID                                 : FFFE / BA368D4D-E147-D242-88FC-61654F8C836C
Duration                                 : 1 min 10 s
Bit rate                                 : 192 kb/s
Channel(s)                               : 2 channels
Channel positions                        : Front: L R
Sampling rate                            : 48.0 kHz
Stream size                              : 1.62 MiB (100%)
```

Sample and documentation:
http://www.psdevwiki.com/ps4/Snd0.at9

JeromeMartinez:
https://github.com/MediaArea/MediaInfoLib/pull/915

PR #908, opened 2018-07-11
ATRAC at9

Sami32:
https://github.com/MediaArea/MediaInfoLib/issues/897


JeromeMartinez:
the .at9 file you provided is actually a WAV file, so your patch does not work.
Adapted version at https://github.com/MediaArea/MediaInfoLib/pull/915.

Sami32:
Hehe, i had suspicion on that, but that was the only sample i could get ;)

Issue #898, opened 2018-07-04
Regression concerning ATRAC3 format


Sami32:
Using the latest MediaInfo shapshot: (same for 17.03)

```
General
Complete name                            : F:\_DEVELOPPEMENT_TOOLS\!MEDIA_TEST_FILES\AUDIO\WMA_CoverArt\bgm03.at3
Format                                   : Wave
File size                                : 472 KiB
Duration                                 : 59 s 614 ms
Overall bit rate                         : 64.9 kb/s
FileExtension_Invalid                    : act wav

Audio
Format                                   : E923AABF-CB58-4471-A119-FFFA01E4CE62
Codec ID                                 : FFFE / E923AABF-CB58-4471-A119-FFFA01E4CE62
Duration                                 : 59 s 614 ms
Bit rate                                 : 64.0 kb/s
Channel(s)                               : 2 channels
Channel positions                        : Front: L R
Sampling rate                            : 44.1 kHz
Stream size                              : 472 KiB (100%)
```

Using 0.7.99 version: (i didn't tested between these 2 versions)
```
General
Complete name                            : F:\_DEVELOPPEMENT_TOOLS\!MEDIA_TEST_FILES\AUDIO\WMA_CoverArt\bgm03.at3
Format                                   : Wave
File size                                : 472 KiB
Duration                                 : 59 s 614 ms
Overall bit rate                         : 64.9 kb/s
FileExtension_Invalid                    : act wav

Audio
Format                                   : Atrac3
Codec ID                                 : FFFE / 58CB7144-23E9-BFAA-A119-FFFA01E4CE62
Duration                                 : 59 s 614 ms
Bit rate                                 : 64.0 kb/s
Channel(s)                               : 2 channels
Channel positions                        : Front: L R
Sampling rate                            : 44.1 kHz
Compression mode                         : Lossy
Stream size                              : 472 KiB (100%)
```

Sami32:
_FFmpeg_ output recognizing correctly the container and codec format:

```
[NULL @ 081e1000] Opening 'sonateno14op27-2.aa3' for reading
[file @ 081e1bc0] Setting default whitelist 'file,crypto'
[oma @ 081e1000] Format oma probed with size=2048 and score=100
[oma @ 081e1000] id3v2 ver:3 flags:00 len:1014
[oma @ 081e1000] Before avformat_find_stream_info() pos: 1120 bytes read:32768 seeks:0 nb_streams:1
[...]

PR #909, opened 2018-07-11
Atrac fix attempt

Sami32:
See #898 and #908 

Not sure if it is the correct way to handle it or even if this change is correct but it just an attempt ;)


JeromeMartinez:
Is fine.
Any idea about the difference between the 2 UUID (per format)?

JeromeMartinez:
> Any idea about the difference between the 2 UUID (per format)?

Argh, actually my mistake, there is only one UUID, the older one is an UUID with inverted bytes, a mistake on my side and I fixed it some time ago (and I missed that the UUID is somewhere else, so the bug)

Sami32:
Yes, the value has changed in MediaInfo somewhere after 7.99

Sami32:
It is why i wasn't sure what way was correct to fix it.

PR #912, opened 2018-07-13
DSD parsers update

JeromeMartinez:
Including DST stream frame rate and duration


Sami32:
Sorry for having missed that.
That is great 👍

PR #915, opened 2018-07-14
ATRAC9 detection 

JeromeMartinez:
From @Sami32 (adapted from https://github.com/MediaArea/MediaInfoLib/pull/908)


Source/Resource/Text/DataBase/CodecID_Audio_Riff.csv
```diff
@@ -235,9 +235,7 @@ DFAC;FrameServer;;DebugMode SonicFoundry Vegas FrameServer ACM Codec;
 F1AC;FLAC;;Free Lossless Audio Codec;
 FFFE;Extensible;;;
 FFFF;In Development;;;
-58CB7144-23E9-BFAA-A119-FFFA01E4CE62;Atrac3;;;
```
JeromeMartinez: @Sami32 I remove the ones because they are due to buggy display of UUIDs in previous versions of MediaInfo, no more used now (in current version of MediaInfo).
If you don't know the version of MediaInfo used by your users, you have to check the CodecID with both values (the buggy one and the real one) on your side.

Sami32: I was afraid to broke some of your stuff doing that ;)

Thank you 😃

Sami32: But is same for Atrac9, as i set the UID from 7.99 and latest SNAPSHOT, so one have to be removed as well.

Issue #594, opened 2017-08-21
FLAC is not correctly identified in MOV container


Sami32:
FLAC is identified as 19D in MOV container (having FFV1 video). FFmpeg identify it correctly.

Sample taken from here:
https://archive.org/download/test20100622_v210_and_ffv1/DHC36.ffv1.flac.mov

Sami32:
Btw, while generate an ISMV file from this one with the following command line:
`ffmpeg -y -i DHC36.ffv1.flac.mov -f ismv testout.mp4`
or
`ffmpeg -y -i DHC36.ffv1.flac.mov testout.ismv`

As already reported here:
https://github.com/MediaArea/MediaInfo/issues/173

JeromeMartinez:
Fixed by https://github.com/MediaArea/MediaInfoLib/pull/916.

PR #916, opened 2018-07-14
WAV: Support of CodecID 19D

JeromeMartinez:
fix for https://github.com/MediaArea/MediaInfoLib/issues/594 (is a MOV but use WAV CodecIDs)
@Sami32


Sami32:
I thought about that yesterday but as i was not sure my fix attempt was the correct approach i didn't go on and waited feedback first.

Thanks :)

PR #924, opened 2018-07-16
oga ogv ogx

Sami32:
This https://wiki.xiph.org/Ogg link could also have been choose but i found the Wikipedia one clear and well done.


JeromeMartinez:
Please update the PR with the same modification in File__Analyze_Automatic.cpp

Sami32:
Ok, finally Github status have began to recover...
https://github.com/MediaArea/MediaInfo/pull/266

PR #930, opened 2018-08-03
Support of injection of external metadata

JeromeMartinez:


JeromeMartinez:
@dericed please compile and test this version.

a.csv should contain e.g.:
```
Audio;0;X;Y
Audio;0;X2;Y2
```

command:
```
./mediainfo FileName --ExternalMetadata=file://a.csv --Output=PBCore2
```

PBCore2 new lines:
```
                <essenceTrackAnnotation annotationType="X">Y</essenceTrackAnnotation>
                <essenceTrackAnnotation annotationType="X2">Y2</essenceTrackAnnotation>
```

By default external metadata line is ignored if there is same metadata already in the file.
append `;1` (5th column is a bool "replace instead of ignoring") to the line for using external metadata in priority.

JeromeMartinez:
> This looks interesting.

say thanks to @dericed :).

> will this be a command line option eventually, or will it use some API to inject another schema into a mediainfo output?

command line & API, see my previous comment for example.

JeromeMartinez:
CSV with ";" as column separator (1 day I'll use the standard CSV, not the one from Excel in Europe...)
Columns:
1: stream kind (General, Video, Audio, Text...)
2: 0-based stream number
3: field name
4: field value
5 (optional): replace instead of ignoring if field is already present (metadata from the file)

If only 1 line (no carriage return), can be directly provided to the command line instead of using a file.

dericed:
I think the majority of such use would be for "General;0". Perhaps if argument list is less than 4 than presume first two would be "General;0". For instance so "General;0;x;y" would be equivalent to "x;y".

JeromeMartinez:
command:
```
./mediainfo FileName "--ExternalMetadata=File_Created_Date;Y" --Output=PBCore2
```
(note: works also with text file in case you need more than 1 field; a second "--ExternalMetadata" would just erase the first one)

PBCore2 new lines:
```
       <instantiationAnnotation annotationType="File_Created_Date">Y</instantiationAnnotation>
```

please test again.

kieranjol:
This looks absolutely wonderful,and something I never thought would be possible with mediainfo.

JeromeMartinez:
@dericed I added the support of several instances of the option on the command line.
Still one instance if e.g. you reference a file on disk with several lines for the metadata input
"--ExternalMetadata=" (empty metadata) removes previous metadata (may be useful with some automation scripts)

PR #921, opened 2018-07-15
Fix #913

pavel-pimenov:
https://sourceforge.net/p/mediainfo/bugs/1022/

Issue #913, opened 2018-07-13
Segfault during MediaInfo_Open_Buffer_Finalize


frebib:
Running on Alpine Linux which uses musl libc instead of glibc.
This crash is pretty repeatable, with pretty much any video file.

```
Thread 8 "Threadpool work" received signal SIGSEGV, Segmentation fault.
decfloat (f=f@entry=0x7ffff1f1aeb8, c=50, bits=24, emin=-149, sign=1, pok=pok@entry=1) at src/internal/floatscan.c:67
67	{
(gdb) bt
#0  decfloat (f=f@entry=0x7ffff1f1aeb8, c=50, bits=24, emin=-149, sign=1, pok=pok@entry=1) at src/internal/floatscan.c:67
#1  0x00007ffff7d8c76e in __floatscan (f=f@entry=0x7ffff1f1aeb8, prec=prec@entry=0, pok=pok@entry=1) at src/internal/floatscan.c:509
#2  0x00007ffff7dbbfd5 in strtox (s=0x555558fb4578 "23.976", p=0x7ffff1f1aff0, prec=prec@entry=0) at src/stdlib/strtod.c:14
#3  0x00007ffff7dbc00c in strtof (s=<optimized out>, p=<optimized out>) at src/stdlib/strtod.c:22
#4  0x00007fffeeb124c0 in void std::__convert_to_v<float>(char const*, float&, std::_Ios_Iostate&, int* const&) () from /usr/lib/libstdc++.so.6
#5  0x00007fffeeb5dadf in std::num_get<wchar_t, std::istreambuf_iterator<wchar_t, std::char_traits<wchar_t> > >::do_get(std::istreambuf_iterator<wchar_t, std::char_traits<wchar_t> >, std::istreambuf_iterator<wchar_t, std::char_traits<wchar_t> >, std::ios_base&, std::_Ios_Iostate&, float&) const () from /usr/lib/libstdc++.so.6
#6  0x00007fffeeb584aa in std::num_get<wchar_t, std::istreambuf_iterator<wchar_t, std::char_traits<wchar_t> > >::get(std::istreambuf_iterator<wchar_t, std::char_traits<wchar_t> >, std::istreambuf_iterator<wchar_t, std::char_traits<wchar_t> >, std::ios_base&, std::_Ios_Iostate&, float&) const () from /usr/lib/libstdc++.so.6
[...]

PR #931, opened 2018-08-09
 Fix of segfault infinite loop

JeromeMartinez:
Due to https://github.com/MediaArea/MediaInfoLib/pull/921 while fixing https://github.com/MediaArea/MediaInfoLib/issues/913.
@pavel-pimenov the test is too much as we have a 0.001 precision. I changed the value, still an ugly fix of an ugly code though.
@frebib I installed Alpine Linux 3.8.0, installed g++/make/zlib-dev and compiled MediaInfo 18.05, but I can not reproduce the segfault, so please test this version.


frebib:
@JeromeMartinez Originally I could not reproduce the segfault with the mediainfo cli on Alpine, but only with libmediainfo when called from Mono. I should be able to confirm this fix in the next couple of days hopefully

JeromeMartinez:
@frebib looks like there is a problem with floating point option somewhere... Would you mind to provide a basic example of Mono program which was crashing? I think it would help me to not reproduce the problem when I change this part of the code in the future.

PR #939, opened 2018-08-10
CarriageReturnReplace option

JeromeMartinez:
Quick and dirty code for being able to disable carriage return replacement when reading metadata.
@Sami32 add MediaInfo::Option("CarriageReturnReplace", "") in your code before calling MediaInfo::Open and you'll get the field as is (without modifications, so containing carriage returns)


Sami32:
👍

PR #941, opened 2018-08-11
Legacy option set to false by default

JeromeMartinez:
Do not output very old & redundant fields used only by very old programs.
` --Legacy=1` if needed by such old program.


JeromeMartinez:
> Is there a list of values

All lines with "Deprecated" word in the description at https://github.com/MediaArea/MediaInfoLib/tree/master/Source/Resource/Text/Stream.

I didn't deprecated things in the last years (last one in 2010 IIRC, I didn't remove for 1 customer I was working with) so I don't think you'll get issues, but right better to double-check.
Anyway, if you are not sure, use the option ;-).

kieranjol:
Brilliant,thank you!

PR #942, opened 2018-08-11
HighestFormat option set to true by default

JeromeMartinez:
Show the highest format detected, instead of the lowest one.
Examples:
- "AAC LC SBR PS" instead of "AAC" because a LC decoder with SBR and PS support is needed for fully decoding it (else only the 24 kHz / 1 channel part is decoded)
- "E-AC-3 JOC" instead of "E-AC-3" because an E-AC-3 decoder with JOC (Atmos) support is needed for fully decoding it (else the Atmos part is ignored)


JeromeMartinez:
Nudge @Sami32 because I know it may impact you.

PR #949, opened 2018-08-14
+ Fix #1005: Obey edit list in QuickTime Timecode track

pchapmanvc:
Obey edit list in QuickTime timecode tracks. Fixes reporting for the sample file in:
https://sourceforge.net/p/mediainfo/bugs/1005/

Signed-off-by: Peter Chapman <peterc@telestream.net>


JeromeMartinez:
Your patch does not handle time code tracks with non continuous time codes i.e. you add an offset based on time offset to the timecode of first frame without checking the timecode at this time offset.
Checking how we could handle such files, we need to check stts atom, find the corresponding chunk (instead of using the first chunk) and use this one instead of the first one, also adapt FirstEditOffset correspondingly. Would you be motivated for implementing it?

If not, I see that 99% of files have continuous time codes (a single time code value) and 99% of the remaining 1% of files have an offset of 0, so I guess it is not a big issue and for these files current behavior is already not correct so this is changing a bug with another bug for these files, I may accept the patch as is, just noticing that this is "only" a partial fix.

pchapmanvc:
Hi Jerome, I see your point, thanks. However:
* I don't have a test file with multiple timecode samples and a non-trivial edit list.
* I think that a correct implementation would require me to spend some more time learning how MediaInfoLib's parsing process works.
At the moment I'm not going to develop the patch any further. I understand why you might not want a 'partial' fix so I'm happy whatever you decide.
Thanks.

Issue #953, opened 2018-08-16
ExternalMetaDataConfig_Get': is not a member of 'MediaInfoLib::MediaInfo_Config'


pavel-pimenov:
error C2039: 'ExternalMetaDataConfig_Get': is not a member of 'MediaInfoLib::MediaInfo_Config'

![default](https://user-images.githubusercontent.com/920705/44186476-4c8bf200-a122-11e8-971f-f10dbf758fa7.png)

Fix:
    #if defined(MEDIAINFO_EBUCORE_YES) || defined(MEDIAINFO_NISO_YES)
?

JeromeMartinez:
ExternalMetadata is now usable in some other cases.
So the opposite, is https://github.com/MediaArea/MediaInfoLib/pull/954 OK?

PR #954, opened 2018-08-16
Fix issue with some #defines

JeromeMartinez:
Fix https://github.com/MediaArea/MediaInfoLib/issues/953


JeromeMartinez:
Oops, I should test it even when I quickly add something... Fixed.

PR #969, opened 2018-08-27
Matroska: show smallest supported version only if corresponding option is set

JeromeMartinez:
Contains an update of https://github.com/MediaArea/MediaInfoLib/pull/968


kieranjol:
Does that mean that you won't see the 'Version 4/ Version 2' type of thing
by default?

JeromeMartinez:
Will be "Version 4" instead.

PR #972, opened 2018-08-29
more pbcore2 updates

dericed:


Source/MediaInfo/Export/Export_PBCore2.cpp
```diff
@@ -408,12 +441,19 @@ Ztring Export_PBCore2::Transform(MediaInfo_Internal &MI, version Version)
     Child->Add_Attribute_IfNotEmpty(MI, Stream_General, 0, General_Format_Version, "annotation");
 
     //formatLocation
-    Node_Main.Add_Child("instantiationLocation", MI.Get(Stream_General, 0, General_CompleteName));
+    if (!MI.Get(Stream_General, 0, __T("instantiationLocation")).empty())
+        Node_Main.Add_Child_IfNotEmpty(MI, Stream_General, 0, "instantiationLocation", "instantiationLocation");
+    else
+        Node_Main.Add_Child("instantiationLocation", MI.Get(Stream_General, 0, General_CompleteName));
```
JeromeMartinez: On one side I don't like the idea of a "if" here, `--ExternalMetadata="CompleteName;bla" ` could be used instead, on other other side I understand you want to use the PBCore2 name.

dericed: In testing, users requested specifically to not share the filepath (privacy) and wanted an option to overwrite that value. We could do `if instantiationLocation or CompleteName`.

JeromeMartinez: I understand the need to remove filepath (we offer the option on [MediaInfoOnline](https://mediaarea.net/MediaInfoOnline) for the same reason), just wondering if it is not better to keep something generic (so you use `--ExternalMetadata="CompleteName;bla" ` and remove this additional code) in order to avoid to increase the already big code when it is not necessary.
In other words, I am not strongly against and I'll accept the PR, just please consider the added value of using `instantiationLocation` explicitly i.e. if you think it is easier fr your user to use such command, to be used only if it is important for you to have it, else just use `CompleteName` in the ExternalMetadata call (note that `CompleteName` works for all formats).

PR #1010, opened 2018-09-25
MPEG2 variable GOP detectin fix

karasjov:
For MPEG2 streams, when a frame is encoded in the form of separate fields, MediaInfo erroneously detects second filed of an I-frame as a beginning of a new GOP. That leads to picture_coding_types being populated with I-frame only GOP consisting of I-frame's first fields and the remainder of the GOP with all but first frames being duplicated. That eventually breaks variable GOP detection.

I suggest considering only the first filed in calculations, in order to achieve expected behavior. In my tests, proposed changes didn't brake neither progressive stream handling, nor interlaced streams with picture structure "frame".


JeromeMartinez:
Good catch, thanks.

Issue #1012, opened 2018-10-02
CodeGuard - Access overrun AC3_BitRate[bit_rate_code]*1000


pavel-pimenov:
https://github.com/MediaArea/MediaInfoLib/blob/300b062529b9ed31cfc8b84c79cb4da1b6187a5e/Source/MediaInfo/Multiple/File_Mpeg_Descriptors.cpp#L2869

------------------------------------------
Error 00025. 0x130600 (Thread 0x12B8):
Access overrun: Attempt to access 2 byte(s) at 0x010B0546+72, that is at
 offset 0+72 in block 0x010B0546(=PreRelease.exe:0x02:00C546) which is only 38
 bytes long.
Call Tree:
   0x00936C84(=PreRelease.exe:0x01:535C84) ..\..\..\Source\MediaInfo\Multiple\File_Mpeg_Descriptors.cpp#2869
   0x00926D9A(=PreRelease.exe:0x01:525D9A) ..\..\..\Source\MediaInfo\Multiple\File_Mpeg_Descriptors.cpp#1391
   0x00560E6B(=PreRelease.exe:0x01:15FE6B) ..\..\..\Source\MediaInfo\File__Analyze.cpp#2359
   0x0055C52C(=PreRelease.exe:0x01:15B52C) ..\..\..\Source\MediaInfo\File__Analyze.cpp#1513
   0x00559FEA(=PreRelease.exe:0x01:158FEA) ..\..\..\Source\MediaInfo\File__Analyze.cpp#1080
   0x005562D9(=PreRelease.exe:0x01:1552D9) ..\..\..\Source\MediaInfo\File__Analyze.cpp#693
   0x00559B42(=PreRelease.exe:0x01:158B42) ..\..\..\Source\MediaInfo\File__Analyze.cpp#1031
   0x004CFF1A(=PreRelease.exe:0x01:0CEF1A) ..\..\..\Source\MediaInfo/File__Analyze.h#96
   0x00968AB3(=PreRelease.exe:0x01:567AB3) ..\..\..\Source\MediaInfo\Multiple\File_Mpeg_Psi.cpp#2290
   0x0095EB0B(=PreRelease.exe:0x01:55DB0B) ..\..\..\Source\MediaInfo\Multiple\File_Mpeg_Psi.cpp#1061
   0x0095D003(=PreRelease.exe:0x01:55C003) ..\..\..\Source\MediaInfo\Multiple\File_Mpeg_Psi.cpp#992
[...]

PR #1014, opened 2018-10-02
Access overrun AC3_BitRate

pavel-pimenov:
#1012

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mpeg_Descriptors.cpp
+++ b/Source/MediaInfo/Multiple/File_Mpeg_Descriptors.cpp
@@ -2864,12 +2864,12 @@ void File_Mpeg_Descriptors::Descriptor_81()
     int8u sample_rate_code, bit_rate_code, surround_mode, bsmod, num_channels, langcod, textlen, text_code;
     bool language_flag, language_flag_2;
     BS_Begin();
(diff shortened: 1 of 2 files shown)


Source/MediaInfo/Multiple/File_Mpeg_Descriptors.cpp
```diff
@@ -2866,7 +2866,7 @@ void File_Mpeg_Descriptors::Descriptor_81()
     BS_Begin();
     Get_S1 (3, sample_rate_code,                                "sample_rate_code"); if (sample_rate_code<4) {Param_Info2(AC3_SamplingRate[sample_rate_code], " Hz");}
     Skip_S1(5,                                                  "bsid");
-    Get_S1 (6, bit_rate_code,                                   "bit_rate_code"); Param_Info2(AC3_BitRate[bit_rate_code]*1000, " Kbps");
+    Get_S1 (6, bit_rate_code,                                   "bit_rate_code"); if (bit_rate_code<19) {Param_Info2(AC3_BitRate[bit_rate_code]*1000, " Kbps");}
```
JeromeMartinez: Not important (no need to edit), but for next time, prefer
Param_Info2C(bit_rate_code<19, AC3_BitRate[bit_rate_code]*1000, " Kbps");
(the test code is deleted by preprocessor in that case, when small package is requested)

PR #1017, opened 2018-10-08
Fix missing zconf.h

bgermann:
The CMake build breaks because zlib's CMake build moves zconf.h to
zconf.h.included. This adds the build directory as include directory.


JeromeMartinez:
Thanks.

PR #1020, opened 2018-10-10
MPEG-4: use ISO 8859-1 locale when showing atom codes

JeromeMartinez:
Using the system locale is faster but fails with e.g. © (not pure ASCII) when the locale is UTF-8 (nowadays all systems except Windows), when showing atom codes in the output e.g. MediaTrace.

TODO: current code for ISO 8859-1 locale is slower but can be accelerated.

Additional fix for https://github.com/MediaArea/MediaConch_SourceCode/issues/630 on non Windows machine.
cc @dericed .


JeromeMartinez:
Superseed.

Issue #875, opened 2018-06-18
Added Matroska tags ignored


mkver:
As you might know, MKVToolNix (or actually mkvpropedit, the tool that is also the backend for the header and the chapter editor of the MKVToolNix GUI) contains a cli option named `--add-track-statistics-tags` which parses the whole file and adds track statistics tags in the same way that mkvmerge does it by default. But MediaInfo does not really use these tags although it could [determine that they are up to date](https://gitlab.com/mbunkus/mkvtoolnix/wikis/Automatic-tag-generation#determining-if-the-tags-are-up-to-date). Note: Not all muxers (including ffmpeg's muxer) write the segment info `date` value that is needed for this check.
Here is the output for such a file. As you can see, Mediainfo's "Bit rate" and "Stream size" fields aren't there; instead the tags are more or less copied verbatim (which is supoptimal for people/programs who parse MediaInfo's output):
```
General
Unique ID                                : 238307966721314686418052194793101143739 (0xB34871E596239DD29F5BBF4189F32EBB)
Complete name                            : I:\Matroska.Tags.mkv
Format                                   : Matroska
Format version                           : Version 4 / Version 2
File size                                : 56.9 KiB
Duration                                 : 4 s 341 ms
Overall bit rate                         : 107 kb/s
Encoded date                             : UTC 2018-06-18 10:10:30
Writing application                      : mkvmerge v24.0.0 ('Beyond The Pale') 64-bit
Writing library                          : libebml v1.3.6 + libmatroska v1.4.9

Video
ID                                       : 1
Format                                   : VP9
Codec ID                                 : V_VP9
Duration                                 : 4 s 338 ms
Width                                    : 1 920 pixels
Height                                   : 1 080 pixels
Display aspect ratio                     : 16:9
Frame rate mode                          : Constant
Frame rate                               : 23.976 (24000/1001) FPS
Language                                 : English
Default                                  : Yes
Forced                                   : No
Statistics Tags Issue                    : mkvpropedit v24.0.0 ('Beyond The Pale') 64-bit 2018-06-18 10:34:27 / mkvmerge v24.0.0 ('Beyond The Pale') 64-bit 2018-06-18 10:10:30
FromStats_BitRate                        : 11285
FromStats_Duration                       : 00:00:04.254708333
FromStats_FrameCount                     : 102
FromStats_StreamSize                     : 6001

Audio
ID                                       : 2
Format                                   : Opus
Codec ID                                 : A_OPUS
[...]

PR #1024, opened 2018-10-13
x I875, Matroska: Fix for added Statistics Tags

mkver:
Hello,

in f1f2d80a9a8995f32e229b8aaade40a460490e40 you tried to solve issue #875 about added Matroska statistics tags that have been ignored, but you got the comparison of the dates wrong so that only tags that are newer than the file would be treated as invalid. You quickly noticed this and fixed it in 8579631710d4c726a7268822aeec84384b5a3b45, but you (most likely unintentionally) reversed this in eb0999b917054020f08daba490a87dcf41be8f2f. This PR is intended to fix this.

Greetings
mkver


JeromeMartinez:
Oops... Tanks for having remarked that and provided the fix, appreciated.

PR #1028, opened 2018-10-20
Small fix

pavel-pimenov:
V808 'Nodes' object of 'vector' type was created but was not utilized. MediaInfo_Inform.cpp 480
V817 It is more efficient to seek '.' character rather than a string. Export_Niso.cpp 79


JeromeMartinez:
thanks!

Issue #1030, opened 2018-10-20
Large internal read request can cause buffer overflow


CatWEax:
If `Config.File_Buffer_Size_ToRead` is set to something large as a result of parsing MKV attachments (issue #1029) or parsing a crafted MKV file, subsequent read requests can overflow the limited buffer at `Config.File_Buffer`.

`Config.File_Buffer` is enlarged for large read requests, but its size is limited to 64 MiB:
https://github.com/MediaArea/MediaInfoLib/blob/3d6927b090bd5c4283c28467b8e45d6b9626874d/Source/MediaInfo/Reader/Reader_File.cpp#L510-L514

When a subsequent read request is issued, the size of the read request does not take the limited buffer size into account:
https://github.com/MediaArea/MediaInfoLib/blob/3d6927b090bd5c4283c28467b8e45d6b9626874d/Source/MediaInfo/Reader/Reader_File.cpp#L612

`Partial_End` can exceed input file size:
https://github.com/MediaArea/MediaInfoLib/blob/3d6927b090bd5c4283c28467b8e45d6b9626874d/Source/MediaInfo/Reader/Reader_File.cpp#L277

If `Config.File_Buffer_Size_ToRead` exceeds EOF and `Partial_End` exceeds file size, the read size is limited to the remainder of the file. If the remainder of the file is larger than 64 MiB, `Config.File_Buffer` will overflow.

On Linux, this overflow is detected by libc and the application (`mediainfo`) is killed due to *double free or corruption*. Someone else tested it for me on Windows and it didn’t crash.

JeromeMartinez:
I remarked that too while fixing the other issue, working on that.

JeromeMartinez:
(and I definitely need to refactor this part of the code :( )

JeromeMartinez:
@CatWEax I was fixing in the Matroska parser, but actually it is better to fix it in Reader_File too in case this issue appears elsewhere.
Your analysis of the source of the issue is appreciated.
So 2 PR: 1 for this issue as a general issue, and anotehr one will come for Matroska specific issue.

PR #1031, opened 2018-10-20
Reader_File: Fix crash in case of request to read a big count of bytes

JeromeMartinez:


JeromeMartinez:
Fix https://github.com/MediaArea/MediaInfoLib/issues/1030

PR #1074, opened 2019-01-26
Fix File_Mxf AcquisitionMetadata leaks

Ozomahtli:


JeromeMartinez:
We try to support old platforms as much as possible, and some of them still don't appreciate C++11.
Thanks for https://github.com/MediaArea/MediaInfoLib/pull/1075.

PR #1075, opened 2019-01-27
Fix File_Mxf AcquisitionMetadata Leaks, C++98

Ozomahtli:


JeromeMartinez:
Thanks for detection of the leak + corresponding fix!

PR #1105, opened 2019-03-28
'mdhd_TimeScale' variable is checked after it was used.

pavel-pimenov:
V781 Perhaps there is a mistake in program logic. Check lines: 2023, 2035. File_Mpeg4.cpp 2023


JeromeMartinez:
I am not convinced by the error message, it is weird, but the patch is useful (division by 0), thanks.

PR #1106, opened 2019-04-08
Typos in language file

JeromeMartinez:


Source/MediaInfo/File_Dummy.cpp
```diff
@@ -165,7 +165,7 @@ void File_Dummy::Fill_Dummy_General()
     Fill(Stream_General, 0, General_ExecutiveProducer, "ExecutiveProducer");
     Fill(Stream_General, 0, General_ProductionDesigner, "ProductionDesigner");
     Fill(Stream_General, 0, General_CostumeDesigner, "CostumeDesigner");
-    Fill(Stream_General, 0, General_Choregrapher, "Choregrapher");
+    Fill(Stream_General, 0, General_Choreographer, "Choregrapher");
```
Nadahar: @JeromeMartinez I don't know anything about your changes, I just noticed that the string still says ´Choregrapher` even thought the identifier has been corrected to `Choreographer`. Is that intentional?

JeromeMartinez: It is intentional, in a global change of the typo everywhere, from https://github.com/MediaArea/MediaInfo/pull/219.
Hoping that it will break nothing...

JeromeMartinez: (previous post deleted, I saw my mistake)

No, it was not intentional. It does not hurt a lot as it is for the template used when testing, but better to fix it everywhere.

Thanks!

PR #1150, opened 2019-08-02
* Fix C++Builder 10.x

pavel-pimenov:
[bcc32 Error] File__Analyze.h(492): E2015 Ambiguity between 'MediaInfoLib::Char' and 'ZenLib::Char'
 [bcc32 Warning] File_Usac.cpp(921): W8004 'nDrcChannelGroups' is assigned a value that is never used


JeromeMartinez:
Thanks!

PR #1166, opened 2019-09-02
chore(mediainfo): Escape extended ASCII chars

XMegamanX:
Fix Mediaconch error: [220](https://github.com/MediaArea/MediaConch/issues/220) and [220](https://github.com/MediaArea/MediaConch/issues/220) and Mediaconch_sourceCode [658](https://github.com/MediaArea/MediaConch_SourceCode/issues/658)

### DIFF
--- a/Source/MediaInfo/File__Analyze_Element.cpp
+++ b/Source/MediaInfo/File__Analyze_Element.cpp
@@ -364,6 +364,8 @@ static size_t Xml_Content_Escape_MustEscape(const char* Content, size_t Size)
             default:
                 if (C<0x20)
                     return Pos;
+                if (C>0x7F)
+                    return Pos;
         }
     }
 


JeromeMartinez:
The patch proposal is too much aggressive: it transforms any non ASCII string, so also valid UTF-8 string, in binary dump (in base64). This is not what is intended.
It would also hide the real source of the issue.

Anyway, your patch helped me to quickly get the source of the issue.
Please try with [this change](https://github.com/JeromeMartinez/MediaInfoLib/commit/83a35bda09d57107b1f2095735a88913211298d9).
FYI, this issue comes from "component name" field in MOV files, I was redirecting the raw string, which has sometimes accents to the output, so ASCII chars >=0x80 were not converted to UTF-8 as expected, [this commit](https://github.com/JeromeMartinez/MediaInfoLib/commit/0a162b7bf2d7e59e546872fbd48b41ec793758c3) is an attempt to better handle all scenarios I found in my sample files (muxers developers are very creative when they need to store this field... and non standard).

If it also provide a fix on your side, I'll close this PR in favor of a PR with the 2 cited commits.

XMegamanX:
Hello ! Hell yeah, that's a good news. Yeah, I know that it was a little bit too aggressive but it was the less worst proposal I can give after having a look at the complex source code between MediaConch and MediaInfoLib. Thank you for your time. I will wait for your PR to get a proper commit version to download.
Do you know when it will be released/merged ?

Regards,

Megaman

JeromeMartinez:
PR deprecated by #1168

> Do you know when it will be released/merged ?

Merged just after you say it is fine for you.
then in snapshots of MediaConch, I am not sure I want to do a release of MediaConch "only" for that, and for the moment MediaConch release are very rare.

PR #1180, opened 2019-11-07
add mov discrete audio channel description definitions for Discrete-16 to Discrete-31

jaystevens:
This patch adds audio channel description definitions for Discrete-16 to Discrete-31 for Quicktime movies.
Post Production regularly makes 24+ channel Quicktime files for network delivery with discrete labeled audio.


JeromeMartinez:
Thanks. I guess I should have something smarter for 1-31 stuff, but it is fine enough for the moment.

PR #1182, opened 2019-11-11
Nullptr exception when there is no vui in some mp4 files

stanislavpetkov:
Nullptr exception when there is no h264 vui in some mp4 files, which have DTVCC embedded in h264 SEI message.


JeromeMartinez:
Thanks!

Issue #1181, opened 2019-11-09
crash MediaInfoLib::File_Avc::Streams_Fill


pavel-pimenov:
File 
Pink.Floyd.Shine.On.You.Crazy.Diamond.Part.1.Concert.Screen.Reels_by.Gionick2014.mkv

1. dockerfile

FROM centos

RUN yum install -y git automake autoconf libtool pkgconfig make gcc gcc-c++ zlib-devel

RUN git clone https://github.com/MediaArea/ZenLib.git \
    && cd ZenLib/Project/GNU/Library \
    && ./autogen.sh \
    && ./configure --enable-static \
    && make

RUN git clone https://github.com/MediaArea/MediaInfoLib.git \
    && cd MediaInfoLib/Project/GNU/Library \
    && ./autogen.sh \
    && ./configure --enable-static --enable-staticlibs \
    && make

RUN git clone https://github.com/MediaArea/MediaInfo.git \
    && cd MediaInfo/Project/GNU/CLI \
    && ./autogen.sh \
    && ./configure --enable-staticlibs \
    && make \
    && make install

WORKDIR /MediaInfo
RUN ls
RUN mediainfo  --Help
COPY 1.mkv 1.mkv
RUN ls
RUN yum install -y valgrind
RUN valgrind mediainfo /MediaInfo/1.mkv

2. crash

Step 12/12 : RUN valgrind mediainfo /MediaInfo/1.mkv
 ---> Running in 50c61b3cd075
==1== Memcheck, a memory error detector
==1== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==1== Using Valgrind-3.14.0 and LibVEX; rerun with -h for copyright info
==1== Command: mediainfo /MediaInfo/1.mkv
==1==
==1== Invalid read of size 4
==1==    at 0x7C277C: MediaInfoLib::File_Avc::Streams_Fill(__gnu_cxx::__normal_iterator<MediaInfoLib::File_Avc::seq_parameter_set_struct**, std::vector<MediaInfoLib::File_Avc::seq_parameter_set_struct*, std::allocator<MediaInfoLib::File_Avc::seq_parameter_set_struct*> > >) (in /usr/local/bin/mediainfo)
==1==    by 0x7C3849: MediaInfoLib::File_Avc::Streams_Fill() (in /usr/local/bin/mediainfo)
[...]

PR #1189, opened 2019-11-14
File_Mk - fix crash

pavel-pimenov:
https://github.com/MediaArea/MediaInfoLib/issues/1181
  https://github.com/MediaArea/MediaInfo/issues/371


JeromeMartinez:
Apologizes for the (huge) delay, but it is merged now (slightly modified version).
Thanks a lot for having caught this ugly piece of code.

PR #1193, opened 2019-11-20
RPM & Arch python packages

g-maxime:

### DIFF
--- a/Project/GNU/PKGBUILD
+++ b/Project/GNU/PKGBUILD
@@ -2,15 +2,15 @@
 # Maintainer: MediaArea.net SARL <info@mediaarea.net>
 # Contributor: hydro <hydro@freenet.de>
 
-pkgname=libmediainfo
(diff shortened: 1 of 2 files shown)


Project/GNU/PKGBUILD
```diff
@@ -2,15 +2,15 @@
 # Maintainer: MediaArea.net SARL <info@mediaarea.net>
 # Contributor: hydro <hydro@freenet.de>
 
-pkgname=libmediainfo
+pkgname=('libmediainfo' 'python2-mediainfodll' 'python-mediainfodll')
```
JeromeMartinez: I think I prefer `python-mediainfo`, using `dll` here may be misleading (Windows thing) and this name comes from Debian IIRC, maybe not good to reproduce the same thing in other distros.

g-maxime: Done

PR #1201, opened 2019-12-17
Fix x variable was assigned the same value.

pavel-pimenov:


Source/MediaInfo/Video/File_Vc3.cpp
```diff
@@ -710,7 +710,7 @@ void File_Vc3::Data_Parse()
         }
         else
         {
-            FrameInfo.PTS=FrameInfo.DTS=FrameInfo.DUR=(int64u)-1;
```
JeromeMartinez: Why? Not same assignment (PTS, DTS, DUR are different)

pavel-pimenov: error :-(

pavel-pimenov: revert

PR #1207, opened 2020-01-10
Update parameter definitions

ablwr:
This PR is for filling out and clarifying values in MediaInfo's `--Info-Parameters` page. 

Now ready for review!

### DIFF
--- a/Source/Resource/Text/Stream/Audio.csv
+++ b/Source/Resource/Text/Stream/Audio.csv
@@ -1,192 +1,192 @@
-Count;;;N NI;;;Count of objects available in this stream
-Status;;;N NI;;;bit field (0=IsAccepted, 1=IsFilled, 2=IsUpdated, 3=IsFinished)
-StreamCount;;;N NI;;;Count of streams of that kind available
-StreamKind;Audio;;N NT;;;Stream type name;;
-StreamKind/String;;;N NT;;;Stream type name
-StreamKindID;;;N NI;;;Number of the stream (base=0)
-StreamKindPos;;;N NI;;;When multiple streams, number of the stream (base=1)
-StreamOrder;;;N YIY;;;Stream order in the file, whatever is the kind of stream (base=0)
-FirstPacketOrder;;;N YIY;;;Order of the first fully decodable packet met in the file, whatever is the kind of stream (base=0)
-Inform;;;N NT;;;Last **Inform** call
-ID;;;N YTY;;;The ID for this stream in this file
-ID/String;;;Y NT;;;The ID for this stream in this file
-OriginalSourceMedium_ID;;;N YTY;;;The ID for this stream in the original medium of the material
-OriginalSourceMedium_ID/String;;;Y NT;;;The ID for this stream in the original medium of the material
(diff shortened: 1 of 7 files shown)


ablwr:
I have been moving through these parameters (more than what is in this PR currently) and am trying to keep questions to a minimum, but I still have a few meta-questions. Can you tell me if these assumptions are correct?

1. `Count`, `Status`, and `Inform` all seem like they are primarily used for internal purposes, is that true?

2. For the syntax between the name and definition, the pattern like "N YTY", I am assuming Letter[0] is Y for default display, N for not (only show on `-f`), then a space, and then another string. I'm assuming T for Text, I for Integer, F for Float, and the extra Ys and Ns are whether to show the field at all if there's no value. Is that kinda right?


3. "Original" is used to mean the data coming from the codec and not the container, right?

4. I found the definition and meaning of "Source" [on Sourceforge](https://sourceforge.net/p/mediainfo/feature-requests/366/#f4cf), but does that apply to new fields (that don't seem to be in use) like `MasteringDisplay_ColorPrimaries_Source`? What do you mean when you use "Source"?

5. What is your definition of "Nominal" for bit rates and frame rates? I am working with some competing definitions more broadly.

ablwr:
Something else I cannot figure out in the Video section is the meaning of these four fields:
`Stored_Width,  Stored_Height and `Sampled_Width, Sampled_Height`? They don't have any current definitions and it's not clear from looking at the code or other resources.

ablwr:
:innocent:

JeromeMartinez:
sorry about the delay, it is a lot to review... And still late on hot stuff.
I don't forget!

ablwr:
Is this still blocked?

JeromeMartinez:
Super long delay but it will be merged soon!
Rebased, slightly edited.

ablwr:
WOW!!!!!!!!!!!!!!!!!! :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada:

ablwr:
Thank you, this is great! Can't wait to check it out in the next release!!!!

JeromeMartinez:
> Can't wait to check it out in the next release!!!!

hmmm... Yes and no. It will be in the next release in the source code, but the binary delivery [will be without the descriptions](https://github.com/MediaArea/MediaInfoLib/pull/1712) (there will be a hint for checking the official repo at https://github.com/MediaArea/MediaInfoLib/tree/master/Source/Resource/Text/Stream , the idea is not to hide the description).

We are working on reducing the binary size, especially for the JavaScript and the Amazon Lambda targets, bigger binary means slower loading time, and theses additions were inflating the binary size much (not a criticism, it is good to have theses descriptions).

ablwr:
Ahh got it, I saw those PRs come in earlier today. That's good. 

Are there plans to display params in a UX-friendlier way on the MediaArea site?

JeromeMartinez:
> Are there plans to display params in a UX-friendlier way on the MediaArea site?

We should definitely do that for more visibility, something similar to https://mediaarea.net/en/MediaInfo/Support/Tags, but we are so laaaaate on website update (the page I linked is so old, need a big refresh).
Please add a ticket on https://github.com/MediaArea/MediaArea-Website/issues for tracking that.

PR #1209, opened 2020-01-24
CMake: Copy source files into binary dir

1480c1:
Same as https://github.com/MediaArea/ZenLib/pull/105

A work around for certain compilers not working on windows due to extra long paths (Something like `CMakeFiles/mediainfo.dir/D_/actions/_work/media-autobuild_suite/media-autobuild_suite/build/libmediainfo-git/Source/ThirdParty/aes-gladman/aes_modes.c.o` under `D:/actions/_work/media-autobuild_suite/media-autobuild_suite/build/libmediainfo-git/build`)

Since it the source files aren't subdirectories, cmake uses absolute paths to the files when deciding the output name.

### DIFF
--- a/Project/CMake/CMakeLists.txt
+++ b/Project/CMake/CMakeLists.txt
@@ -93,8 +93,13 @@ elseif(MSVC)
 else()
   set(MediaInfoLib_CONFIG_INSTALL_DIR "${LIB_INSTALL_DIR}/cmake/mediainfolib")
 endif()
-
-set(MediaInfoLib_SOURCES_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../../Source)
+if(MINGW)
+  # Work around for cmake generating extra long paths
+  file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/../../Source DESTINATION .)
+  set(MediaInfoLib_SOURCES_PATH ${CMAKE_CURRENT_BINARY_DIR}/Source)
+else()
+  set(MediaInfoLib_SOURCES_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../../Source)
+endif()
 
 # use bundled tinyxml only if no system
 find_package(TinyXML)
@@ -392,6 +397,10 @@ set(MediaInfoLib_SRCS
 if(NOT TinyXML_FOUND)
   include_directories(${MediaInfoLib_SOURCES_PATH}/ThirdParty/tinyxml2/)
   list(APPEND MediaInfoLib_SRCS ${MediaInfoLib_SOURCES_PATH}/ThirdParty/tinyxml2/tinyxml2.cpp)
+endif()
+
+if(MINGW)
+  set_source_files_properties(${MediaInfoLib_SRCS} PROPERTIES GENERATED true)
 endif()
 
 add_library(mediainfo ${MediaInfoLib_SRCS})


JeromeMartinez:
While working on CMake, I saw that I forgot this one :(. Still relevant?
@cjee21 as you are on CMake things, what do you think about it?

cjee21:
I am not sure, seems only affect mingw and when path is too long like using the autobuild_suite if I remember correctly from the ZenLib PR. I also remember there was a mistake at ZenLib there where it check mingw when copy files but check WIN32 when clean files, resulting in clean command deleting the original source files on MSVC instead of the intention of only deleting the copied temp source files for mingw. That I fixed in ZenLib quite long ago.

cjee21:
I think can merge this but change

```
if(WIN32)
```

to match the other added condition above (MINGW) before merging. Same bug as the ZenLib one which will cause original source files to be deleted on clean with other non-mingw Windows compilers.

cjee21 (approved):

cjee21:
By the way, this manual copying may not rebuild if the original sources changes and user does not reconfigure right? Also may not rebuild properly without reconfigure if it was cleaned (file commands supposedly only run during configure). Is there a better solution? Does https://cmake.org/cmake/help/latest/variable/CMAKE_OBJECT_PATH_MAX.html help in this case?

cjee21:
Is any workaround even necessary? I cannot reproduce any compile issues using mingw on Windows. When the path gets too long and reach the limit, CMake automatically converts them to a hash and mingw still can build it successfully without any warnings/errors and produces `libmediainfo.a`.

`D:\actions\_work\012345678901234567890123456789\012345678901234567890123456789\build\libmediainfo-git\build\CMakeFiles\mediainfo.dir\d2b701a4b22464bd5e7cb8e25eed6f34\build\libmediainfo-git\Source\ThirdParty\aes-gladman\aes_modes.c.obj`

`D:\actions\_work\012345678901234567890123456789\012345678901234567890123456789\build\libmediainfo-git\build\CMakeFiles\mediainfo.dir\f089ab42e28eefe7d1e82cb50af97a05\Source\MediaInfo\Multiple\File__ReferenceFilesHelper_Resource.cpp.obj`

cjee21:
> Also may not rebuild properly without reconfigure if it was cleaned

Confirmed this workaround at ZenLib breaks `cmake --build . --clean-first` with mingw. Removing the similar workaround from ZenLib fixed it.

1480c1:
I don't think it's needed anymore since it was created at a time when the
260 path limit was more relevant, but not easily fixed.

cjee21:
> I don't think it's needed anymore since it was created at a time when the
260 path limit was more relevant, but not easily fixed.

So @JeromeMartinez, it's not needed anymore. CMake should automatically hash the paths as needed to keep them under 260 chars. If there are any compilers that can't handle 260 chars or some rare case where it fails to keep it under 260, the first solution should be to set a lower `CMAKE_OBJECT_PATH_MAX` (min 128 char).

https://github.com/MediaArea/ZenLib/commit/243a5d154f5a0dde95872500e9635c8ad41044f9 and https://github.com/MediaArea/ZenLib/commit/65b9152e19a58a16a4ec671dea8a8f23292dadb6 should be reverted since it is no longer needed and breaks build after clean with mingw as well as not rebuilding when source files are updated and no longer in sync with the manual copy.

JeromeMartinez:
OK, I was too late on that :).
So I close this one and will remove the commits on ZenLib.

PR #1210, opened 2020-01-30
AVC slice header - chroma weight table fix

3CHosler:
File_Avc.cpp::pred_weight_table
Currently there is only one set of chroma weight/offset
values read per list(l0, l1), after all the luma coefficient sets.
But I believe there should be two sets for each set of luma
coefficients.

Noticed when mediatrace returned a value of -355 for
slice_beta_offset_div2, which has range (-6,+6), and 21 for
disable_deblocking_filter_idc in range (0,+2). 😄


JeromeMartinez:
> But I believe there should be two sets for each set of luma coefficients.

Oops, I checked the specs, right.

> Noticed when mediatrace [...]

Happy to see that [MediaTrace](https://mediaarea.net/MediaTrace) is useful!

PR #1218, opened 2020-03-02
Some 128-bit identifiers well not well displayed in multi-thread

JeromeMartinez:
Fix https://github.com/sbraz/pymediainfo/issues/76#issuecomment-593392976.


sbraz:
That works as expected and returns the right ID, thanks :)

PR #1249, opened 2020-04-23
support timecode hours up to +/-99 before resorting to mod 24

### LINKED ISSUE #1248: Supporting timecode of 24:00:00:00 upwards in MOV/MP4 timecode tracks
Hello,

I've come across an interesting MOV file which has a timecode track indicating start time of `24:00:00:00`.

In ffprobe, the start timecode is output as `24:00:00:00`, however in MediaInfoLib it is reported as `00:00:00:00`.

I have narrowed the difference down to the following change I have tried:

https://github.com/vectronic/MediaInfoLib/commit/8dd12dd1a5d0921fa5bc12fc47198d82a107234d

I realise this change is not usable in it's current form as it doesn't account for numbers greater than 99 taking up more than 2 digits in timecode format.

I have looked into the specs for mov/mp4 and can't see anything which says the timecode can't be `24:00:00:00` or greater and I can see `24HourMax=No`.

With the above change the timecode tracks outputs as:

```
Time code of first frame                 : 24:00:00:00
```

One thing I have not been able to solve is that even with the above change, the start time code for the video track is still output as: 

```
Time code of first frame                 : 00:00:00:00
```

I believe this should refer to the timecode track?

It looks like at some point this value is ignored despite the following appearing in the MediaInfoLib full output for the video track:

```
Delay                                    : 86400000
Delay                                    : 24 h 0 min
Delay                                    : 24 h 0 min 0 s 0 ms
Delay                                    : 24 h 0 min
Delay                                    : 24:00:00.000
```

and the following in the traces (where the DTS is calculated based on the timecode):

```
00004672   slice_start - Frame (decoding order) 0 - Frame (presentation order) 2 - picture_coding_type I - temporal_reference 2 - DTS 24:00:00.000 - time_code 0:0:0.80 (2178 bytes)
```

In summary I guess I am asking:

* Is the limiting to timecode below 24h intentional? If not, would you be happy for a PR with better handling of hours above 99?
* Am I on the wrong path thinking that the video start time code should also display `24:00:00.000` ? If I am not wrong, do you have any pointers to where a fix could be made for this?

I have attached full output of the file before and after the change as well as a full parsing trace output (without the change).

[example-output.txt](https://github.com/MediaArea/MediaInfoLib/files/4523695/example-output.txt)

[example-output-after-change.txt](https://github.com/MediaArea/MediaInfoLib/files/4523694/example-output-after-change.txt)

[mediainfo-trace.txt](https://github.com/MediaArea/MediaInfoLib/files/4523585/mediainfo-trace.txt)

vectronic:
fixes #1248

### DIFF
--- a/Source/MediaInfo/File__Analyze.h
+++ b/Source/MediaInfo/File__Analyze.h
@@ -1344,7 +1344,6 @@ protected :
     friend class File__Tags_Helper;
     friend class File_Mk;
     friend class File_Mpeg4;
-    friend class File_Mk;
 
     //***************************************************************************
     // Helpers


JeromeMartinez:
Would you mind [squashing commits](https://stackoverflow.com/questions/27988573/flatten-commit-history-using-tortoisegit)?

PR #1263, opened 2020-05-14
Matroska: support of AVC and HEVC CodecPrivate extensions

JeromeMartinez:


quietvoid:
There are now available samples of the new implemented Matroska spec, if it's useful for this feature: https://www.makemkv.com/download/dvtest/

PR #1272, opened 2020-06-04
MKV/WebM: Report Delay and Duration correctly for VP9 MKV/WebM files.

MaxEliaserAWS:
No parser had been implemented for the VP9 codec, and the MKV analyzer was
not reporting Delay for any codec without a parser.

Since the parser is not actually necessary to extract the accurate Delay for
MKV video files, I did not fix this by implementing a VP9 parser; rather I
allowed the MKV analyzer to report a Delay if no parser is present.

With the Delay value detected correctly, the Duration can now be calculated
correctly as well.

This test video is an excerpt from [Big Buck Bunny](https://peach.blender.org/about/) which is copyright 2008, [Blender Foundation](https://www.blender.org/foundation/) and licensed under [CC BY 3.0](http://creativecommons.org/licenses/by/3.0/).
[test_video_000000002.zip](https://github.com/MediaArea/MediaInfoLib/files/4732455/test_video_000000002.zip)

Because the test video has a nonzero initial timestamp, its duration will not be reported correctly unless the Delay is detected. This patch fixes the duration in the Video section of the mediainfo report.

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mk.cpp
+++ b/Source/MediaInfo/Multiple/File_Mk.cpp
@@ -1148,49 +1148,49 @@ void File_Mk::Streams_Finish()
             }
         }
 
-        if (Temp->second.Parser)
+        //Delay
+        if (Temp->second.TimeCode_Start!=(int64u)-1 && TimecodeScale)
         {
-            //Delay
-            if (Temp->second.TimeCode_Start!=(int64u)-1 && TimecodeScale)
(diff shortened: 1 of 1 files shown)


MaxEliaserAWS:
Most of the diff chatter is just indentation changes. Recommend hiding whitespace to review.

MaxEliaserAWS:
Hi Jerome, glad I could help!

PR #1281, opened 2020-06-23
Remove UTF-8 BOM from metadata if present

JeromeMartinez:
Useless and potential XML output validation issue.
Also reformat if BOM indicates a wrong order of bytes.
Referenced in https://github.com/sbraz/pymediainfo/issues/90#issuecomment-646915113.


sbraz:
I just tested this branch with https://github.com/sbraz/pymediainfo/files/4807132/sample.zip and I see an empty copyright. Am I missing something?

JeromeMartinez:
> I see an empty copyright. 

You are on a non-Windows platform and I was working on a fix while you wrote :-p.
It should be fine now (with the updated commit).

sbraz:
Works fine now, thanks :)

PR #1294, opened 2020-09-14
MXF, add FFV1 support

JeromeMartinez:
Based on an early draft, not to be used in prod.


kieranjol:
Do such files exist?

JeromeMartinez:
> Do such files exist?

As far as I know there is 2 existing files with that for the moment, created by another party based a draft spec we are working on together (hello [Open LossLess in STanDards project](https://mediaarea.net/ollistd) :) ), I prefer to avoid to share them publicly for the moment because it is an early stage, MXF ULs may change.

JeromeMartinez:
> Do such files exist?

@kieranjol [MXF/FFV1 sample files](https://www.digitizationguidelines.gov/guidelines/MXF_app_sampleFiles.html#2022).

PR #1299, opened 2020-09-21
Dolby E: better trace

JeromeMartinez:
Long time ago, I received the [Dolby E High-Level Frame Description](https://developer.dolby.com/globalassets/professional/dolby-e/dolby-e-high-level-frame-description.pdf) with an NDA as the doc was not yet online, so I was not able to use the exact field names and much metadata parsing. Now the doc is online so the NDA does not apply :-p, and as I was working on Dolby E, I prefer to have only one code and merge code I used in a private project into MediaInfo open source.
Nothing fancy here, more for [MediaTrace](https://mediaarea.net/MediaTrace), but later e.g. peak_meter or AC-3 related metadata could be exported, as well as CRC checks.

& better handling of bit rate (both the total encoded content and estimation per program).


ValZapod:
Why it recognises all sample rates as 48000? FFmpeg actually specifies and actually uses those (and 48000 does not even EXIST, WTF). https://github.com/FFmpeg/FFmpeg/commit/44e27d937d88f3e7cdc83f4de1b9d3a590bcaa04#diff-ec791aea00ab0f331287a3ffd7765c130b63f9b57d8bc62cd55be8611a5ab29aL85 

I read [here](https://forum.videohelp.com/threads/400100-Dolby-E-Decoding-FFmpeg) that 48000 should be forced anyway while actually speeding it up? 

I need it for Eurovision rip, can you help?

JeromeMartinez:
> Why it recognises all sample rates as 48000?

As far as we know, Dolby E is only 48 kHz. I am interested in a stream if you think it is not 48 kHz.
"fr_code" means "frame rate code", I don't get the reason FFmpeg translates it to sample rate, on our side we provide the sample rate (48 kHz, hard coded) and the frame rate (24, 25, 29.97, 30... based on this "fr_code"), 2 separate fields, of a Dolby E Stream.

ValZapod:
> our side we provide the sample rate and the fr_code

FFmpeg uses fr_code too. The problem is that it needs to produce PTS and DTS and how will it do it with two different synchronisations? sample_rate_tab[header->fr_code] is what it does and writes it to header->sample_rate. Now the problem is I do not quite understand whether conversion to 48000 should be done and obviously only two people know on FFmpeg project.

JeromeMartinez:
> how will it do it with two different synchronisations?

I don't get how they do, I can just say that we sync on frame rate (PTS frame X = PTS frame X-1 + 1/framerate), the count of samples per frame is set accordingly (48000/framerate).
So in short: I don't see any issue in MI about that.

ValZapod:
LOL, there is an actual SMPTE timecode inside more modern stream /facepalm.

metadata_crc, metadata_extension_crc and audio_subsegment1_crc: are all NOKed, does not look like you verify them correctly or it is ffmpeg's fault.

Some audio_extension_subsegment0_crc:     43726 (0x0AACE) - (20 bits) are not even checked at all.

JeromeMartinez:
Right, there are some issues with the CRC check in MI, I need to check what is wrong (it is also the case with files from Dolby). But should not have any impact on the MI report.

ValZapod:
BTW, this patch decodes some samples more correct: 

[dolby_e @ 000002534a95e5c0] Truncated mantissas at 945, highest frequencies not recoverable

https://patchwork.ffmpeg.org/project/ffmpeg/patch/20220913213127.1756-1-nicolas.gaullier@cji.paris/

JeromeMartinez:
> Sampling rate               : 44.8 kHz

Please provide a sample file with such output.

ValZapod:
As an example 
http://0x0.st/oOvv.zip

JeromeMartinez:
> I am still confused about 44800 Hz.

The WAV file says so, see bytes 0x18-0x2A, I can do nothing about that, buggy encoder.
The source Dolby E is 48 kHz.

```
00000C  Stream format - Audio (48 bytes)
00000C   Header (8 bytes)
00000C    Name:                                fmt 
000010    Size:                                40 (0x00000028)
000014   FormatTag:                            65534 (0xFFFE)
000016   Channels:                             8 (0x0008)
000018   SamplesPerSec:                        44800 (0x0000AF00)
00001C   AvgBytesPerSec:                       716800 (0x000AF000)
000020   BlockAlign:                           16 (0x0010)
000022   BitsPerSample:                        16 (0x0010)
000024   cbSize:                               22 (0x0016)
000026   ValidBitsPerSample / SamplesPerBlock: 16 (0x0010)
000028   ChannelMask:                          1599 (0x0000063F)
00002C   SubFormat:                            00000001-0000-0010-8000-00AA00389B71
```

ValZapod:
I am talking about indu_ex.s337m file, when you press ffplay indu_ex.s337m it prints 44800.

>I can do nothing about that, buggy encoder.

Decoder you mean?

JeromeMartinez:
> Decoder you mean?

Either the Dolby E decoder or the WAV ~encoder~ muxer (I was thinking to the WAV muxer).

> I am talking about indu_ex.s337m file, when you press ffplay indu_ex.s337m it prints 44800.

Now I have doubts.
AFAIK, Dolby E is always 48 kHz, only the frame rate may change.
Or I am wrong there... I don't decode the stream, I just do Dolby E --> 48 kHz.
I don't see in the [DolbyE High-Level Frame Description](https://developer.dolby.com/globalassets/professional/dolby-e/dolby-e-high-level-frame-description.pdf) doc some tip about it.

You made me curious, I check the (non public) Dolby E SDK, doc says SDK output 48 kHz, but... also that there is up or downsampling! A Dolby E frame is always 1792 samples, your file is at 25 fps so 1792 * 25 = 44800 Hz.

I was too quick, they are right, because they don't do up or downsampling.
I should do that (showing stored sampling rate) too. (but fun, for e.g. 29.97 fps content, sampling rate is not an integer: 1792 * 30 / 1.001 = 53706,24 Hz)

ValZapod:
>is just that FFmpeg can not handle non-integer frame rate.

This is bad :(

PR #1303, opened 2020-09-26
Add dependency_pid in DOVI descriptor

jcdr428:
With dual PID profile 7 streams, the DOVI descriptor can include the dependency_pid field. 

Cf. dolby-vision-bitstreams-in-mpeg-2-transport-stream-multiplex-v1.2.pdf:

Dovi


JeromeMartinez:
Please combine the commits in one.
Would it be useful to show the dependency_pid in the MediaInfo report?

jcdr428:
@JeromeMartinez done in #1304 .
Mediainfo will show both video tracks/PIDs and PID 0x1015 is always deemed to depend on PID 0x1011. 
Not sure it would be very useful to show the dependency_pid in Mediainfo.

JeromeMartinez:
FYI you can force push on your branch instead of creating a new PR.
Dependency ID may be nice to have, but could be done later, and I merged the other PR, thanks.

PR #1308, opened 2020-09-30
Recommended setting of CMAKE_MODULE_PATH

tw-git:
A more flexible, more generic and recommended setting of CMAKE_MODULE_PATH. See examples here:
https://cgold.readthedocs.io/en/latest/tutorials/cmake-sources/includes.html#modify-correct
https://cgold.readthedocs.io/en/latest/tutorials/cmake-sources/includes.html#modify-incorrect
With this change, parent projects are able to propagate custom modules to MediaInfoLib, like a custom FindZenLib.cmake file.

### DIFF
--- a/Project/CMake/CMakeLists.txt
+++ b/Project/CMake/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 2.8.11)
 
 project(MediaInfoLib)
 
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
Nothing against that, @g-maxime fine for you?
by the way, thank you for checking our (ugly) build script, don't hesitate to suggest more improvements.

g-maxime (approved):

PR #1318, opened 2020-10-27
Crashes

g-maxime:


Source/MediaInfo/Multiple/File_Mpeg4.cpp
```diff
@@ -425,7 +426,7 @@ void File_Mpeg4::Streams_Finish()
     if (Retrieve(Stream_General, 0, General_Format)==__T("Final Cut EIA-608"))
     {
         for (streams::iterator Stream=Streams.begin(); Stream!=Streams.end(); ++Stream)
-            for (size_t Pos=0; Pos<Streams[(int32u)Element_Code].Parsers.size(); Pos++)
+            for (size_t Pos=0; Pos<Streams[(int32u)Element_Code].Parsers.size() && Pos<Stream->second.Parsers.size(); Pos++)
```
JeromeMartinez, line 429: Actually the main issue is the mismatch between `Streams[(int32u)Element_Code]` and `Stream`, I edit this commit in order to remove the mismatch.

Source/MediaInfo/Video/File_Hevc.cpp
```diff
@@ -3245,7 +3245,7 @@ void File_Hevc::VPS_SPS_PPS()
             Element_Begin1("nalUnit");
             int16u nalUnitLength;
             Get_B2 (nalUnitLength,                              "nalUnitLength");
-            if (nalUnitLength<2 || Element_Offset+nalUnitLength>Element_Size)
+            if (nalUnitLength<2 || Element_Offset+nalUnitLength>Element_Size || Element_Offset+nalUnitLength>BS->Remain())
```
JeromeMartinez, line 3248: `BS->Remain()` is in bits and actually not relevant here, it leads to breaks in the analysis of some HEVC streams. I remove this commit from the PR.

PR #1325, opened 2020-11-10
Add support for TIFF compression codes 7 and 8

david-russo:
These were defined in the [Adobe Photoshop Technical Notes from 2002](https://www.adobe.io/content/dam/udp/en/open/standards/tiff/TIFFphotoshop.pdf), a link to which has also been included. The document outlines Deflate compression as well as a better method for using JPEG compression in TIFFs than was previously defined in the TIFF v6 specification (which was released prior to JPEG's ISO standardization).

Also, keep up the great work. It's much appreciated.


JeromeMartinez:
Thanks! A bit afraid that it may break some automatic parsers expecting "JPEG", anyway it would not be the first time and it is not the biggest usage, so OK.

PR #1336, opened 2020-12-02
Fix crash in ffv1 parser

g-maxime:


JeromeMartinez:
This part needs a complete rewrite but the patch is fair enough as a quick fix, thanks.

PR #1350, opened 2021-01-21
Remove URL parameters from FileName

g-maxime:


JeromeMartinez:
I think I prefer to use `Reader_libcurl_FileNameWithoutPassword()` everywhere (also in the error logs), possible to use it with the additional filter on "?"?

g-maxime:
@JeromeMartinez I can chain the two functions call, or add a parameter to Reader_libcurl_FileNameWithoutPassword()

JeromeMartinez:
add a parameter to Reader_libcurl_FileNameWithoutPassword(), I consider the part after ? as like a password (it is actually the case for AWS presigned URLs!).

g-maxime:
So I delete the parameters also from the logs?

JeromeMartinez:
Yes

PR #1352, opened 2021-01-28
INT_MAX/MIN used where INT8_MAX/MIN should be

LeoDavidson:
A few places were assigning INT_MIN or INT_MAX to A.obj_gain_db when it's only 8-bits wide and is documented as using INT8_MIN and INT8_MAX as its special values.

Most places in the code use the correct constants, but three were missing the "8".


JeromeMartinez:
Thanks!

PR #1360, opened 2021-02-10
Add support for passing vectors of paths into MediaInfoList

hjmallon:
This adds an `Open` call for a vector of file paths.


cjee21:
Looks inefficient, creating a vector of one element when opening with a string. What about calling the existing string open in a loop for the vector one instead. Existing apps would be looping the string call when opening multiple files so this change will cause creation of many vectors with single element. The vector call cannot be exposed in the C API so not all apps can migrate to vector call too.

PR #1368, opened 2021-02-17
Fix parsing of WM/EncodingTime

LeoDavidson:
The Windows Media parsing code tried to handle "WM/EncodingTime" but was failing because the value is a FILETIME, which uses units of 100 nanoseconds from 1601, but it was passing the value as-is to something that expects seconds from 1601. Just needs a division by 10000000 to fix.

I found a side-effect of fixing this which could be a breaking change for dependencies:

- File_Wm::Header_FileProperties sets Stream_General / General_Encoded_Date to the CreationDate from the header.

- With this fix in place, File_Wm::Header_ExtendedContentDescription also sets Stream_General / General_Encoded_Date to the "WM/EncodingTime" date-time (if the file has the tag).

- This results in MI.Get("Encoded_Date") returning a string with two date-times. For example:
   "UTC 2005-11-03 16:04:06.505 / UTC 2021-01-01 14:20:24"
   The first is the time from the header, and the second is WM/EncodingTime.

That's fine for me, as I just made my code parse the last date-time in that case. But maybe it will break code that doesn't expect anything extra after the seconds or optional milliseconds of the first date-time.

If you want to test this, you can set WM/EncodingTime on a test WMV file via Windows 10's Properties dialog. The field is, confusingly, called "Media Created", at least in English.

I can make and share a small test file if it would be useful and save you time.

### DIFF
--- a/Source/MediaInfo/Multiple/File_Wm_Elements.cpp
+++ b/Source/MediaInfo/Multiple/File_Wm_Elements.cpp
@@ -265,7 +265,10 @@ void File_Wm::Header_FileProperties()
     //Filling
     if (MaximumBitRate)
         Fill(Stream_General, 0, General_OverallBitRate_Maximum, MaximumBitRate);
-    Fill(Stream_General, 0, General_Encoded_Date, Ztring().Date_From_Milliseconds_1601(CreationDate/10000));
+    Ztring Encoded_Date_New=Ztring().Date_From_Seconds_1601(CreationDate/10000000);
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
Ouch... Thanks for catching the bad code + the explanation + the fix.

I modified the PR:
- Removed the extra merge commit
- Removed the full description in the commit
- Added some code for coherency between the 2 sources (all in seconds, and no duplicate if values from the 2 sources of metadata are same; modification by Windows explorer will show 2 different values, only one if the file is not modified )

Feel free to force push with only your name and digital signature if you wish.

PR #1374, opened 2021-03-03
Use File__Analyze HDR metadata formatting in File__Mxf

hjmallon:
This change unifies the HDR10 metadata handling in Mxf to use the same code as used already by other formats (MP4, Mk etc). Having them separate could introduce differences between the string representations (e.g. "DCI P3" primaries were not recognised in MXF but were in MP4 before this change.

### DIFF
--- a/Source/MediaInfo/File__Analyze.h
+++ b/Source/MediaInfo/File__Analyze.h
@@ -885,7 +885,7 @@ public :
     #if defined(MEDIAINFO_HEVC_YES) || defined(MEDIAINFO_MPEG4_YES)
     void Get_MasteringDisplayColorVolume(Ztring &MasteringDisplay_ColorPrimaries, Ztring &MasteringDisplay_Luminance);
     #endif
-    #if defined(MEDIAINFO_HEVC_YES) || defined(MEDIAINFO_MPEG4_YES) || defined(MEDIAINFO_MATROSKA_YES)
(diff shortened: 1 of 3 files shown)


JeromeMartinez:
>  unifies 

There is definitely a need to refactorize and unify lot of parts of MediaInfo... Thanks for the change, I'll review it soon.

> DCI P3" primaries were not recognised in MXF

Could you share a file demonstrating this issue?

hjmallon:
The MXF code is here (references BT.709, BT.2020 and Display P3)

https://github.com/MediaArea/MediaInfoLib/blob/cba57f58f10ca054a612894ba5843ab187a9d027/Source/MediaInfo/Multiple/File_Mxf.cpp#L3591-L3595

The Generic code is here (references BT.709, BT.2020, Display P3 and DCI-P3.

https://github.com/MediaArea/MediaInfoLib/blob/cba57f58f10ca054a612894ba5843ab187a9d027/Source/MediaInfo/File__Analyze_Streams.cpp#L78-L84

I attach an example file for that specific issue (but I think it is likely that it is not the only example of divergences between the two similar areas of code.

[mediainfo.mxf.zip](https://github.com/MediaArea/MediaInfoLib/files/6075481/mediainfo.mxf.zip)

JeromeMartinez:
Thanks!

PR #1377, opened 2021-03-07
DvDif: Absolute track number + Blank flag, update

JeromeMartinez:
https://github.com/MediaArea/MediaInfoLib/pull/1365 creates some false positives with super buggy files.
Trying to avoid them

- Handle Absolute track number with steps of 5 (current frame start value in the middle of the values from previous frames)
- Handle Absolute track number with steps of 15 (2 then 1 pattern for each Dseq)
- Handle Absolute track number with "random" values in the up to 40 places found; this is [heuristic](https://en.wikipedia.org/wiki/Heuristic)
- Handle negative Absolute track number


dericed (approved):
let's go :)

PR #1380, opened 2021-03-21
File_Mk 'MasteringMetadata': local variable is initialized but not referenced

pavel-pimenov:
warning C4189: 'MasteringMetadata': local variable is initialized but not referenced


JeromeMartinez:
Thanks.

Issue #1252, opened 2020-05-04
Lossless JPEG detected as Lossy


MarcusJohnson91:
Using the latest version of the Mac version of MediaInfo from the app store.

You can see the results in this [Image](https://i.imgur.com/dN2WUXx.png)

Lossless JPEGs use the following Start of Frame (SOF) markers: SOF3, SOF7, SOF11, SOF15 aka 0xFFC3, 0xFFC7, 0xFFCB, 0xFFCF, respectively.

the sof marker number is the only way to truly tell if an image is lossy or lossless.

the image that was misdetected as lossy can be found [here](https://www.dropbox.com/s/tzg5hi9uje6hl3u/1.jpg)

ValZapod:
This one uses SOF3. And it is recognized, just not propagated: https://github.com/MediaArea/MediaInfoLib/blob/2f3998512ec3cc6f55c0905eb993bfce95a9a3b2/Source/MediaInfo/Image/File_Jpeg.cpp#L643

> SOF3, SOF7, SOF11, SOF15

Last time I checked it should not be zeroes, but Os (Start of). Sigh. https://www.itu.int/rec/T-REC-T.81 and https://www.w3.org/Graphics/JPEG/itu-t81.pdf


SOF3 is Lossless (sequential), Huffman coding

SOF11: Lossless (sequential), arithmetic coding

SOF7: differential lossless process frame marker, Huffman coding

SOF15: differential lossless process frame marker, arithmetic coding

JeromeMartinez:
Thanks, I fix that when I have some free time.

ValZapod:
I can give a pull. 0 --> O in two files and add those new markers. Wait a second.

JeromeMartinez:
> I can give a pull

go :)

ValZapod:
BTW, the image as reported as not conformant by reference implementation. Here is my compile with VS 2022.
https://github.com/thorfdbg/libjpeg/issues/61#issuecomment-1034869582

>reading a JPEG file failed - error -1038 - invalid stream, found invalid huffman code in entropy coded segment

JeromeMartinez:
In [snapshots](https://mediaarea.net/download/snapshots/binary/).

PR #1382, opened 2021-03-21
JPEG: zeros in markers are actually letters

ValZapod:
As to how should I do lossless JPEG, I dunno.

Again, lossless JPEG are markers  

SOF3 is Lossless (sequential), Huffman coding

SOF11: Lossless (sequential), arithmetic coding

SOF7: differential lossless process frame marker, Huffman coding

SOF15: differential lossless process frame marker, arithmetic coding

### DIFF
--- a/Source/MediaInfo/Image/File_Jpeg.cpp
+++ b/Source/MediaInfo/Image/File_Jpeg.cpp
@@ -69,22 +69,22 @@ namespace Elements
     const int16u SOP =0xFF91; //JPEG 2000
     const int16u EPH =0xFF92; //JPEG 2000
     const int16u SOD =0xFF93; //JPEG 2000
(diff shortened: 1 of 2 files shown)


JeromeMartinez:
I was not remembering I had the name of markers in the code, i is written "lossless" and I don't check them, oops...
Will merge this PR and will adapt the code for catching lossless.

ValZapod:
FFmpeg also has it, BTW, it even tries to decode SOF3 sample from #1252, but fails, LOL. 

https://github.com/FFmpeg/FFmpeg/blob/781fc06f8e096d273d56cffb6a182da23f96e3c2/libavcodec/mjpegdec.c#L2522

ValZapod:
Also JPEG-LS is signaled by SOF48.

PR #1383, opened 2021-03-21
SMPTE ST 337: support of AC-4

JeromeMartinez:


ValZapod:
Why only case 24 ? Why not all others... https://youtu.be/Ku2w_840BvY?t=1253

JeromeMartinez:
This PR is for AC-4 in SMPTE 337, and AFAIK it is only case 24.

Other cases are either:
- already handled
- not yet implemented in MediaInfo due to lack of file + lack of sponsorship (or volunteer).

I am interested in files for all others not yet supported by MediaInfo.

ValZapod:
> and AFAIK it is only case 24.

Yes. 

I do not have any other samples.

PR #1385, opened 2021-03-24
Add report creation timestamp to text output if requested

g-maxime:


JeromeMartinez:
Argh, merged two quickly, sponsor reported that there is an offset of 1 hour (and this is not local time, it is in the opposite direction). @g-maxime please an update with the right UTC time.

PR #1397, opened 2021-04-14
Fix stream type on encrypted audio mp4

hjmallon:
Encrypted audio files were not having their codec correctly filled as this line is using the wrong StreamKind


JeromeMartinez:
Oops... thanks!

PR #1411, opened 2021-06-27
MP4/QuickTime: Android slow motion real frame rate

JeromeMartinez:
(Unmodified) Android stores high frame rate streams with a "normal" 30 fps stream + MP4 metadata about the real (capture) frame rate.
We show both play back frame rate and real frame rate:

~~~
Frame rate mode                          : Variable
Frame rate                               : 30.045 FPS
Minimum frame rate                       : 28.644 FPS
Maximum frame rate                       : 33.100 FPS
Real frame rate                          : 120.000 FPS
~~~


steveveach:
Excellent feature addition!

PR #1413, opened 2021-07-14
MPEG-4: read too much data with some predecessor definitions

JeromeMartinez:
Fix https://github.com/buzz/mediainfo.js/issues/75.


kernchen:
Thanks @JeromeMartinez  for the prompt action on this one!

PR #1416, opened 2021-08-17
DVRescue: info if non consecutive rectime is less than previous one

JeromeMartinez:


JeromeMartinez:
Fix https://github.com/mipops/dvrescue/pull/340#issuecomment-955574373.
Needs https://github.com/MediaArea/MediaInfoLib/pull/1455.

PR #1423, opened 2021-09-09
Add Dolby ED2 graph output

g-maxime:


JeromeMartinez:
It is difficult to understand well the links when there are alternative layouts in ED2, but not easy to find a good method. Fair enough for now, we may update in the future.

PR #1434, opened 2021-09-22
ADM: compile-in tfsxml unconditionally (unrelated to tinyxml2)

jpalus:
Fixes unresolved symbols when compiling with system tinyxml2.


JeromeMartinez:
Oops... You're right! Thank you.

PR #1442, opened 2021-10-19
Fix mp4 edts trace info

### LINKED ISSUE #1441: mp4 edts media time info is incorrect
1. media_time is a signed value, either `int(64) media_time` or `int(32) media_time`. There is a special value -1. MediaInfoLib handled it as unsigned:
```
name="Media time">4294967295
```

2. The unit of media_time and segment_duration is different:
> segment_duration is an integer that specifies the duration of this edit segment in units of the
timescale in the Movie Header Box
> media_time is an integer containing the starting time within the media of this edit segment (in
media time scale units, in composition time).

Current code use `moov_mvhd_TimeScale` for media_time, which should be `Streams[moov_trak_tkhd_TrackID].mdhd_TimeScale`. However, since mdhd comes after edts, we don't know mdhd_TimeScale yet when handle edts. The following patch doesn't work:

```
diff --git a/Source/MediaInfo/Multiple/File_Mpeg4_Elements.cpp b/Source/MediaInfo/Multiple/File_Mpeg4_Elements.cpp
index 89f6ae0d0..51ddf0e74 100644
--- a/Source/MediaInfo/Multiple/File_Mpeg4_Elements.cpp
+++ b/Source/MediaInfo/Multiple/File_Mpeg4_Elements.cpp
@@ -3856,7 +3856,7 @@ void File_Mpeg4::moov_trak_edts_elst()
         stream::edts_struct edts;
         Element_Begin1("Entry");
         Get_B_DEPENDOFVERSION(edts.Duration,                    "Track duration"); Param_Info2C(moov_mvhd_TimeScale, (int64u)edts.Duration*1000/moov_mvhd_TimeScale, " ms");
-        Get_B_DEPENDOFVERSION(edts.Delay,                       "Media time"); Param_Info2C(moov_mvhd_TimeScale && (edts.Delay!=(int32u)-1), (int64u)edts.Delay*1000/moov_mvhd_TimeScale, " ms");
+        Get_B_DEPENDOFVERSION(edts.Delay,                       "Media time"); Param_Info2C(Streams[moov_trak_tkhd_TrackID].mdhd_TimeScale && (edts.Delay!=(int32u)-1), (int64u)edts.Delay*1000/Streams[moov_trak_tkhd_TrackID].mdhd_TimeScale, " ms");^M
         Get_B4 (edts.Rate,                                      "Media rate"); Param_Info1(((float)edts.Rate)/0x10000);
         Element_End0();
```

quink-black:
1. Media time is signed value
2. Media time is in unit of mdhd time scale. Since mdhd comes after
edts, we don't know mdhd_TimeScale yet when handle edts. Give a hint
to user on which time scale to use.

Fix #1441

### DIFF
--- a/Source/MediaInfo/File__Analyze.h
+++ b/Source/MediaInfo/File__Analyze.h
@@ -574,10 +574,12 @@ public :
     void Get_B2   (int16u  &Info, const char* Name);
     void Get_B3   (int32u  &Info, const char* Name);
     void Get_B4   (int32u  &Info, const char* Name);
+    void Get_B4S  (int32s  &Info, const char* Name);
     void Get_B5   (int64u  &Info, const char* Name);
     void Get_B6   (int64u  &Info, const char* Name);
     void Get_B7   (int64u  &Info, const char* Name);
     void Get_B8   (int64u  &Info, const char* Name);
+    void Get_B8S  (int64s  &Info, const char* Name);
     void Get_B16  (int128u &Info, const char* Name);
     void Get_BF2  (float32 &Info, const char* Name);
     void Get_BF4  (float32 &Info, const char* Name);
--- a/Source/MediaInfo/File__Analyze_Buffer.cpp
+++ b/Source/MediaInfo/File__Analyze_Buffer.cpp
@@ -188,6 +188,16 @@ void File__Analyze::Get_B4(int32u &Info, const char* Name)
 }
 
 //---------------------------------------------------------------------------
+void File__Analyze::Get_B4S(int32s &Info, const char* Name)
+{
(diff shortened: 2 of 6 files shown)


ValZapod:
> need a complete refactoring as well as adding more boxes in the parsing...

No one done it in full, maybe GPAC (but Dolby Vision using GPAC is not seekable on LG C9, so IMHO also bad). EAC3 in mp4 requires both values in editlist and also requires sgpd same as in AAC (decoding from -1 sample). Also apparently non-IDR recovery points as keyframes of AVC need sgpd too!

As for ffmpeg, see this:

ValZapod:
BTW, is this going to break delay relative to video stuff?

PR #1451, opened 2021-11-01
Update C# bitwise Enums to be treated as bitfields

nCubed:
The C# example [HowToUse_Dll.cs](https://github.com/MediaArea/MediaInfoLib/blob/master/Source/Example/HowToUse_Dll.cs) is treating the `Status` Enum as a flagged Enum. Although the example technically works, the Enums should be decorated with the Flags attribute to get the full support of flagged Enums in C#. Looks like the `InfoFileOptions` is a bitwise Enum as well, so decorating with the `Flags` attribute, too.


JeromeMartinez:
Thanks.

PR #1459, opened 2021-11-24
MKV: Include "Encoding settings" from ENCODER_SETTINGS tag

WeebDataHoarder:
Without this changeset, the ENCODER_SETTINGS tag would not display if set in Matroska tags, vs. setting it into h264/h265 stream. This allows setting this flag on bitstreams that do not allow such tags, for example, AV1.
Entry will display as `Encoding settings` instead of `ENCODER_SETTINGS` and be able to use language system.

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mk.cpp
+++ b/Source/MediaInfo/Multiple/File_Mk.cpp
@@ -3195,6 +3195,7 @@ void File_Mk::Segment_Tags_Tag_SimpleTag_TagString()
     if (Segment_Tag_SimpleTag_TagNames[0]==__T("DESCRIPTION")) Segment_Tag_SimpleTag_TagNames[0]=__T("Description");
     if (Segment_Tag_SimpleTag_TagNames[0]==__T("ENCODED_BY")) Segment_Tag_SimpleTag_TagNames[0]=__T("EncodedBy");
     if (Segment_Tag_SimpleTag_TagNames[0]==__T("ENCODER")) Segment_Tag_SimpleTag_TagNames[0]=__T("Encoded_Library");
+    if (Segment_Tag_SimpleTag_TagNames[0]==__T("ENCODER_SETTINGS")) Segment_Tag_SimpleTag_TagNames[0]=__T("Encoded_Library_Settings");
     if (Segment_Tag_SimpleTag_TagNames[0]==__T("FPS")) return; //Useless
     if (Segment_Tag_SimpleTag_TagNames[0]==__T("HANDLER_NAME"))
     {


JeromeMartinez:
Could you provide a sample file?

WeebDataHoarder:
Sure, give me a few minutes to make a smaller one.

WeebDataHoarder:
[video-test.zip](https://github.com/MediaArea/MediaInfoLib/files/7598211/video-test.zip)

## Before:
```
General
Unique ID                                : 31637727357411116664958541704867023765 (0x17CD3463E079024EF3A751BBEBDF0B95)
Complete name                            : video-test.mkv
Format                                   : Matroska
Format version                           : Version 4
File size                                : 756 Bytes
Duration                                 : 120 ms
Overall bit rate                         : 50.4 kb/s
Writing application                      : Lavf58.76.100
Writing library                          : Lavf58.76.100
ErrorDetectionType                       : Per level 1

Video
Format                                   : AV1
Format/Info                              : AOMedia Video 1
Format profile                           : Main@L3.1
Codec ID                                 : V_AV1
Duration                                 : 120 ms
Width                                    : 1 280 pixels
Height                                   : 720 pixels
Display aspect ratio                     : 16:9
Frame rate mode                          : Constant
Frame rate                               : 25.000 FPS
Color space                              : YUV
Chroma subsampling                       : 4:2:0
Writing library                          : Lavc58.134.100 libaom-av1
Default                                  : Yes
Forced                                   : No
Color range                              : Limited
ENCODER_SETTINGS                         : test encoder settings
```

## After:
```
General
Unique ID                                : 31637727357411116664958541704867023765 (0x17CD3463E079024EF3A751BBEBDF0B95)
Complete name                            : video-test.mkv
Format                                   : Matroska
Format version                           : Version 4
File size                                : 756 Bytes
Duration                                 : 120 ms
Overall bit rate                         : 50.4 kb/s
Writing application                      : Lavf58.76.100
Writing library                          : Lavf58.76.100
ErrorDetectionType                       : Per level 1

Video
Format                                   : AV1
Format/Info                              : AOMedia Video 1
Format profile                           : Main@L3.1
Codec ID                                 : V_AV1
Duration                                 : 120 ms
Width                                    : 1 280 pixels
Height                                   : 720 pixels
Display aspect ratio                     : 16:9
Frame rate mode                          : Constant
Frame rate                               : 25.000 FPS
Color space                              : YUV
Chroma subsampling                       : 4:2:0
Writing library                          : Lavc58.134.100 libaom-av1
Encoding settings                        : test encoder settings
Default                                  : Yes
Forced                                   : No
Color range                              : Limited
```

JeromeMartinez:
> Encoding settings                        : test encoder settings

A bit fake :).
Is there a real use case i.e. an encoder/muxer which does it?
Not that I am against the patch, I'll merge it as it makes sense, just that I prefer to have real use cases in my regression tests when I can.

WeebDataHoarder:
Use case: Recently made an AV1 encode where wanted to add encoding settings as x264/x265 do on their bitstream. They do not get added as part of AV1 bitstream, so instead added them while muxing. [Matroska specifies the use of some flags](https://www.matroska.org/technical/tagging.html) including `ENCODER_SETTINGS` so it fits :)

Given this option is more generic than depending on per-codec fields, might be helpful for anyone working with other formats or when the specific encoder they use do not set their own internal flags. Haven't found more precise examples, but seems some people in forums have used the tags as well!

PR #1474, opened 2022-01-13
DV: Handle FSC value even if some blocks are corrupted

JeromeMartinez:
Check FSC value of all blocks of a frame instead of the last header block available. It permits to accept corrupted blocks.
If there are too many corrupted blocks FSC value is considered unknown.

We should do same for other items e.g. FSP, for another PR.

Fix https://github.com/mipops/dvrescue/issues/424.

### DIFF
--- a/Source/MediaInfo/Multiple/File_DvDif.cpp
+++ b/Source/MediaInfo/Multiple/File_DvDif.cpp
@@ -275,6 +275,8 @@ File_DvDif::File_DvDif()
     Speed_FrameCount_Stts_Fluctuation=0;
     Speed_FrameCount_system[0]=0;
     Speed_FrameCount_system[1]=0;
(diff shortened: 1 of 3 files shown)


iamdamosuzuki:
This is awesome, thank you! can't wait to test

JeromeMartinez:
@iamdamosuzuki please test [this snapshot](https://mediaarea.net/download/snapshots/binary/dvrescue/20220114/).

JeromeMartinez:
@iamdamosuzuki please test [this snapshot](https://mediaarea.net/download/snapshots/binary/dvrescue/20220113/).

(sorry, mistake in the issue number, I wanted to mean https://github.com/mipops/dvrescue/issues/424, I handle the other ticket soon)

PR #1487, opened 2022-01-28
Fix off-by-one issue when parsing EXR header flags

### LINKED ISSUE #1486: EXR parser looks at wrong bits in header
Creating a new issue as #490, where I initially wrote about this, is closed.

I'm having trouble running MediaInfo against a EXR file produced by Unreal Engine 4. The issue seems to be that it has long tag names, like `Unreal Camera Final Image Dof Diaphragm Blade Count` - but (according to MediaInfoLib) the LongName flag is **not** set.

Digging a bit deeper into this, as well as trying some other tools against this file, resulted in discovering that MediaInfoLib most likely is mapping wrong bits to the header flags.

[This document](https://openexr.readthedocs.io/en/latest/OpenEXRFileLayout.html#version-field) specifies that the 4 bytes after the magic number is the `version` field, and can be treated as two separate bit fields. Bits 0 - 7 is the version number, bit 9 is the single tile bit, bit 10 is the long name bit, bit 11 is the non-image bit and bit 12 is the multipart bit. **Note that bit 8 is unused.**

`exiftool`, which seems to read the Unreal Engine 4 file correct, follows the above specification: https://github.com/exiftool/exiftool/blob/master/lib/Image/ExifTool/OpenEXR.pm#L172

MediaInfoLib on the other hand seems to have missed the fact that bit 8 is unused, and have a off-by-one issue when reading these flags: https://github.com/MediaArea/MediaInfoLib/blob/master/Source/MediaInfo/Image/File_Exr.cpp#L120

gurka:
Fixes #1486

### DIFF
--- a/Source/MediaInfo/Image/File_Exr.cpp
+++ b/Source/MediaInfo/Image/File_Exr.cpp
@@ -117,11 +117,12 @@ void File_Exr::FileHeader_Parse()
     bool Deep, Multipart;
     Skip_L4(                                                    "Magic number");
     Get_L1 (Version,                                            "Version field");
+    //Note that bit 0 is unused
     Get_L3 (Flags,                                              "Flags");
-        Skip_Flags(Flags, 0,                                    "Single tile");
-        Get_Flags (Flags, 1, LongName,                          "Long name");
-        Get_Flags (Flags, 2, Deep,                              "Non-image");
-        Get_Flags (Flags, 3, Multipart,                         "Multipart");
+        Skip_Flags(Flags, 1,                                    "Single tile");
+        Get_Flags (Flags, 2, LongName,                          "Long name");
+        Get_Flags (Flags, 3, Deep,                              "Non-image");
+        Get_Flags (Flags, 4, Multipart,                         "Multipart");
 
     //Filling
     if (Frame_Count==0)


JeromeMartinez (approved):
Oops, I guess it was working with the other file by luck.
"bit 8-31" then start at "bit 9", so bit 8 is unused, right, LGTM.

Please share a file with such issue so we add it in our regression tests. No need of the content, you can cut the data part if needed.

gurka:
> Oops, I guess it was working with the other file by luck.
Yeah... I tried to download it but it wasn't available anymore.

> Please share a file with such issue so we add it in our regression tests. No need of the content, you can cut the data part if needed.
Will try to find or generate one, or strip the data part as you said as I'm not allowed to share the one I have.

JeromeMartinez:
I checked the other file, actually it was an issue with non long fields names (when field name size is exactly 31 chars), and the code about long field names was actually never tested (only theory). So a real file with that would be nice.

gurka:
Hmm. I were able to generate an EXR using GIMP and then add an attribute with long name using `exrstdattr`. However, the EXR exported from GIMP had bit 9 set (Single-Tile), i.e. the bit the MediaInfoLib reads as LongName flag, so these files can be read OK with MediaInfoLib without this patch.

I guess I need to figure out how to create a EXR with no flags set, and then add a long name attribute so that only 0x400 is set. There are some sample files in https://github.com/AcademySoftwareFoundation/openexr-images that could be used, but I'm not sure if the license for them are OK

JeromeMartinez:
> However, the EXR exported from GIMP had bit 9 set (Single-Tile), i.e. the bit the MediaInfoLib reads as LongName flag, so these files can be read OK with MediaInfoLib without this patch.

this bit is more or less informational only so I am fine for editing it to 0.

> There are some sample files in https://github.com/AcademySoftwareFoundation/openexr-image

In practice some files have the long names flag set but none of them really has a long name.

***

If it is not possible to get a sample file, well we'll keep the old way: not regression test. At least we know it works at least once (you file) now.

gurka:
Ops, I must have mixed up my files - the EXR exported from GIMP has no flags set, so adding an attribute with long name only sets bit 10. Then this file should work: https://github.com/gurka/MediaInfoLib/blob/fix-EXR-header-flags/long-name.exr?raw=true

exiftool 12.39 parses it correctly. MediaInfo(Lib) v21.09 does not. MediaInfo(Lib) with my patch parses it correctly.

JeromeMartinez:
> [...] so adding an attribute with long name only sets bit 10.

Confirmed. With the patch:

```
0000 File Header (8 bytes)
0000  Magic number:                          20000630 (0x01312F76)
0004  Version field:                         2 (0x02)
0005  Flags:                                 4 (0x00000004)
0008  Single tile:                           No
0008  Long name:                             Yes
0008  Non-image:                             No
0008  Multipart:                             No
0008 AVeryLongAttributeNameToTestMediaInfoLib (56 bytes)
0008  Header (52 bytes)
0008   name:                                 AVeryLongAttributeNameToTestMediaInfoLib
0031   type:                                 string
0038   size:                                 4 (0x00000004)
003C  value:                                 (4 bytes)
[...]
```

PR #1498, opened 2022-02-08
E-AC-3: service kind was not correctly handled

JeromeMartinez:
It was OK only for AC-3 or E-AC-3 having a AC-3 core.


JeromeMartinez:
In [Latest snapshots](https://mediaarea.net/download/snapshots/binary/).

PR #1511, opened 2022-03-20
TimeCode: handling of some corner cases

JeromeMartinez:
There are some files with weird config e.g. 25 fps with drop frame time code, trying to have a more coherent output in such cases

### DIFF
--- a/Source/MediaInfo/File__Analyze_Streams.cpp
+++ b/Source/MediaInfo/File__Analyze_Streams.cpp
@@ -2447,15 +2447,17 @@ void File__Analyze::Duration_Duration123(stream_t StreamKind, size_t StreamPos, 
                 FrameRateS=Retrieve(Stream_Audio, StreamPos, "Dolby_Atmos_Metadata AssociatedVideo_FrameRate");
             if (FrameRateS.empty() && StreamKind!=Stream_Audio)
                 FrameRateS=Retrieve(StreamKind, StreamPos, Fill_Parameter(StreamKind, Generic_FrameRate));
+            float64 FrameRateF=FrameRateS.To_float64();
+            int64s FrameRateI=float64_int64s(FrameRateF);
             Ztring FrameCountS;
             if (Parameter==Fill_Parameter(StreamKind, Generic_Duration))
                 FrameCountS=Retrieve(StreamKind, StreamPos, Fill_Parameter(StreamKind, Generic_FrameCount));
             if (FrameCountS.empty() || StreamKind==Stream_Text || (StreamKind==Stream_Audio && Retrieve(Stream_Audio, StreamPos, Stream_Audio)!=__T("PCM")))
             {
                 //FrameCount is not based on frame rate
-                FrameCountS.From_Number(List[Pos].To_float32()*Video_FrameRate_Rounded(FrameRateS.To_float64())/1000, 0);
+                FrameCountS.From_Number(List[Pos].To_float32()*Video_FrameRate_Rounded(FrameRateF)/1000, 0);
             }
-            if (!FrameRateS.empty() && !FrameCountS.empty() && FrameRateS.To_int64u() && FrameRateS.To_int64u()<256)
+            if (!FrameRateS.empty() && !FrameCountS.empty() && FrameRateI && FrameRateI<256)
             {
                 bool DropFrame=false;
                 bool DropFrame_IsValid=false;
@@ -2583,7 +2585,6 @@ void File__Analyze::Duration_Duration123(stream_t StreamKind, size_t StreamPos, 
                 if (!DropFrame_IsValid)
                 {
                     int32s  FrameRateI=float32_int32s(FrameRateS.To_float32());
-                    if (FrameRateI==30 || FrameRateI==60) // Flagging drop frame only for 30/60 DF
                     {
                         float32 FrameRateF=FrameRateS.To_float32();
                         float FrameRateF_Min=((float32)FrameRateI)/((float32)1.002);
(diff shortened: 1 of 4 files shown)


ValZapod:
There is quite strange files that have 120/1.001 fps and that REQUIRE ;100 frames for example (119 frame then 0). Not supported. See for example:
https://drive.google.com/file/d/1cg5iM3HWQZdGP5KR_gxxxWoqSU2mpKDy/view?usp=sharing

Other: 

https://drive.google.com/file/d/1QsJokEmOhTr52ckC4QAk4VvnwZrsF9gV/view?usp=sharing

https://samples.ffmpeg.org/mov/canon_6d/ 

Davinci: https://imgur.com/a/W3SivTD

tcmd spec of mp4 allows that. Will be fixed in ffmpeg here: https://trac.ffmpeg.org/ticket/9710#comment:14

JeromeMartinez:
> There is quite strange files that have 120/1.001 fps and that REQUIRE ;100 frames for example (119 frame then 0). Not supported. 

Clearly not (yet) supported. I get it that it is expected that "number of frames" is for the base 30-fps stuff and that I should multiply, but... How do you indicate that the start frame is 99, for example? When doing that on 120-fps streams it seems that time code can start only on multiples of 4 frames. (I bet that nobody cares because nobody would set a start time code in the middle, but wondering it there is a tip somewhere saying that the frame x at 120 fps is actually the 3rd frame of a theoretical 30 fps stream).

ValZapod:
> How do you indicate that the start frame is 99, for example

The start frame is 100 on that video sample actually. See screenshot in Davinci (it is also V-gamut, -V-log transfer BTW, mpv supports that in -vf=format). The implementation of this is here using variable length of last field. https://patchwork.ffmpeg.org/project/ffmpeg/list/?series=6397

JeromeMartinez:
> The start frame is 100 on that video sample actually. 

I am looking for a file which shows "16:25:24;99" in Davinci, can you provide such file?

> The implementation of this is here using variable length of last field, the frame field. 

this is for displaying the timecode, no problem about that, it is already implemented in my local branch actually.
My concern is about how e.g. "16:25:24;99" is stored. with the patch from FFmpeg I understand that "16:25:24;96" is possible but not ";97", ";98" an ";99", as if it is impossible to cut the file at theses frame numbers. If you cut the file at frame "16:25:24;99", what happens to the time code? it is "16:25:24;99" or "16:25:24;100"? what it the time code of the next frame in the new file?
Also do I understand correctly that after e.g. "16:25:59;119", the next time code is "16:26:00;08" (jump of 2 frames as defined for 30 fps drop frame time code, multiplied by 4 due to 120 fps)?

ValZapod:
> I am looking for a file which shows "16:25:24;99" in Davinci, can you provide such file?

This is 120/1.001 file above. It shows 16:25:24;100 in Davinci. How did you get your value?
SmartSelect_20220416-005251_Gallery

>e.g. "16:25:59;119", the next time code is "16:26:00;08"

The file is too small to get to that.

JeromeMartinez:
> It shows 16:25:24;100 in Davinci. How did you get your value?

I see this one, and I can read out the value if I do as in FFmpeg. I would like to have a file with `16:25:24;99` of `16:25:24;101` as I don't see how it could be written in the file if someone want to cut a file.
Can you for example cut 3 first frames of this file in Davinci and say what timecode is displayed in Davinci and what says FFmpeg?

By looking more deeply in the file, I see that the 30 fps timecode is stored 4x same for each 30 fps values, I bet that Davinci may compute the "intermediate" time code (here: after the cut of 3 first frames "16:25:24;103") should be displayed, I would like confirmation, and from FFmpeg patch it seems that such case is not handled.

ValZapod:
> compute the "intermediate" time code (here: after the cut of 3 first frames "16:25:24;103") should be displayed,

You mean you need to loom into future frames timecode to comput timecode, not only in one frame? Wo.

JeromeMartinez:
> Did you get the ;99 with patch above of ffmpeg?

I **don't** have such file and I **am** looking for one.
I don't have Davinci so I can't cut the file myself.
I would like to know what happens if 3 first frames are removed from the file you provided.
What I am looking for is about how a first timecode of "16:25:24;103" (for example, I need a frame number not a multiple of 4) is stored and if FFmpeg handles it as Davinci (and how).

In summary: please provide a file having a first timecode not a multiple of 4 frames as showed in Davinci. All the provided files have a first timecode being mutliple of 4 so it is not obvious how first timecodes not a multiple of 4 frames are handled by Davinci.

ValZapod:
It is possible not multiple of 4 frames cannot be created. That is why in in XML metadata for ;100 is ;25 (you can see it in mediainfo). Wow, cool. I am not that good in Davinci to lossless create a cut or to change all timecodes, sorry. Free version of davinci cannot do 120 fps too, so you are not in luck, I do have paid version, but I do not know how :(

JeromeMartinez:
> That is why in in XML metadata for ;100 is ;25

Yes, it is also what current MI shows because it is what is written in the file, the x4 is then due to the frame rate indicated.

>  Free version of davinci cannot do 120 fps too, so you are not in luck,

Not easy to cut 3 frames and save the file with a powerful tool? argh :).
I tried to manually edit the time code with [this file](https://store4.gofile.io/download/f382c463-e2f5-459e-9aa5-99c288b42034/P1002234_tc_edited.MOV), what your davinci says? It is just a try, as this is my edit and not a real file.
Then, well, we'll have to wait to someone having Davinci and able to cut 3 frames of a video...

ValZapod:
Nothing changes, I think it reads tmcd, not something in "mdat".

JeromeMartinez:
> Nothing changes, I think it reads tmcd, not something in "mdat".

the tmcd chunk does not have any time code, it is only the "setup" of the timecode in mdat. timecode is either in mdat (time code track) or XML addition. Anyway, we'll see, I adapt MediaInfo code for handling as FFmpeg does, and we'll adapt when someone shows up with a timecode with frame number not a multiple of 4.

ValZapod:
FFmpeg patches applied.

JeromeMartinez:
> There is quite strange file that has 120/1.001 fps and that REQUIREs ;100 frames for example (119 frame then 0). Not supported. 

time codes with frame multiplier due to different frame rate in QuickTime time code track is handled by:
https://github.com/MediaArea/MediaInfoLib/pull/1533

PR #1527, opened 2022-04-13
Fix UWP build

g-maxime:

### DIFF
--- a/Source/MediaInfo/Multiple/File_Nsv.cpp
+++ b/Source/MediaInfo/Multiple/File_Nsv.cpp
@@ -690,6 +690,7 @@ void File_Nsv::Data_Parse()
                                             default            :    Parser=new File_Unknown();
                                         }
                                         break;
+                    default          :  Parser=new File_Unknown();
                 }
                 Open_Buffer_Init(Parser);
                 Stream.Parser=Parser;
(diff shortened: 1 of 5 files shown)


Source/MediaInfo/Multiple/File_Nsv.cpp
```diff
@@ -576,7 +576,7 @@ void File_Nsv::Data_Parse()
 
                 Stream_Prepare(Stream_Type[i]);
                 Fill(Stream_Type[i], 0, Fill_Parameter(Stream_Type[i], Generic_CodecID), Ztring().From_CC4(Stream.codecid));
-                File__Analyze* Parser;
+                File__Analyze* Parser=NULL;
```
JeromeMartinez, line 579: I prefer a `default: Parser=NULL;` at the end of the `switch`.

JeromeMartinez:
If there is a difference of behavior, it may be good to compile the UWP project in the CI.
Or is it "just" a different compilation flag?

Source/MediaInfo/Multiple/File_Nsv.cpp
JeromeMartinez, line 579: I prefer a default: Parser=NULL; at the end of the switch. It will never happen but the the compiler is not aware about it :).

PR #1567, opened 2022-07-17
MOV/MPEG-4: List of QuickTime time code discontinuities, fix

JeromeMartinez:
Fix https://github.com/MediaArea/MediaConch/issues/231#issuecomment-1179644930


dericed:
Any public samples for this?

JeromeMartinez:
> Any public samples for this?

No, but easy to create a fake one based on a "Dave was there" ;-) sample file have non striped time codes.

PR #1572, opened 2022-08-09
Win32: Use non-unicode lib name when unicode is disabled

LAGonauta:
While using newest MediaInfoLib on Windows I wasn't able to compile without Unicode support. 

We must use the non-unicode name as LoadLibrary alias to LoadLibraryA on Windows when _UNICODE is not defined. Therefore LoadLibrary will only accept char*, not wchar_t*.
If we don't do that we get a compilation error when loading the library.


JeromeMartinez:
Thanks.

PR #1575, opened 2022-08-24
MPEG-4/MOV: last time code value for all kind of QuickTime time codes 

JeromeMartinez:
Handle corners cases like edit lists and/or non stripped streams.


JeromeMartinez:
@jsciame @plondino this fix may be useful for you, time code of last frame was not read out (it was computed from time code of first frame + duration, which is fine if there is no discontinuity else is wrong) in case of default parse speed, now it is read out also in case of default parse speed. Not that discontinuities are still not checked in case of default parse speed (still need the option for slower parsing) so you may want the slower parse speed everywhere anyway.

PR #1629, opened 2022-12-01
Display P3 means sRGB transfer

ValZapod:
Mistake in afd79f679ace1b27be429a4eb186c1cd383ca839.

### DIFF
--- a/Source/MediaInfo/Video/File_Mpegv.cpp
+++ b/Source/MediaInfo/Video/File_Mpegv.cpp
@@ -43,8 +43,8 @@ extern const char* Mpegv_colour_primaries(int8u colour_primaries)
         case  8 : return "Generic film";
         case  9 : return "BT.2020";                                     //Added in HEVC
         case 10 : return "XYZ";                                         //Added in HEVC 2014
-        case 11 : return "DCI P3";                                      //Added in HEVC 2016
-        case 12 : return "Display P3";                                  //Added in HEVC 2016
+        case 11 : return "DCI-P3";                                      //Added in HEVC 2016
+        case 12 : return "P3-D65";                                      //Added in HEVC 2016
         case 22 : return "EBU Tech 3213";                               //Added in HEVC 2016
         default : return "";
     }


ValZapod:
@JeromeMartinez Does this break API?

JeromeMartinez:
> Does this break API?

Unfortunately yes, lot of users rely on the exact terms and their scripts will be broken.
I understand the wish to change to [theses names](https://en.wikipedia.org/wiki/DCI-P3#P3_colorimetry) but the "cost" for doing that is too high in my opinion.

ValZapod:
It just that it confuses users a lot. https://forum.doom9.org/showpost.php?p=1979282&postcount=236

Display P3 photos of all iPhones since iPhone 7 is still sdr, and uses pure sRGB (i.e. with linear oart of the curve in ICC profile that uses parametric curve of ICCv4).

But primaries P3-D65 can be combined with PQ. Netflix was doing that for naster JPEG2000 files, until they added BT.2020 primaries a couple of months ago.

ValZapod:
Here is haasn confused about it too. https://github.com/mpv-player/mpv/pull/5236#issuecomment-351904328

:)

JeromeMartinez:
but https://github.com/mpv-player/mpv/pull/5236#issuecomment-352147917 ;-).

Not easy, I need to think more about that, changing again and again (you remarked well that I already changed it once) is not good too.

On one side I am reluctant to change again, on another side I like to avoid marketing names like "Display P3". You may convince me at some point...

BTW, "DCI-P3" and not "P3-DCI" in order to have "P3-xxx"? But after that we have issues with "DCI-P3+"...

For my education, could you share sample files with different settings, especially the one with misleading MediaInfo report, and explain what it is + expected MediaInfo report.

ValZapod:
>but https://github.com/mpv-player/mpv/pull/5236#issuecomment-352147917 ;-).

It says the same as I do. 

>This color space uses the DCI P3 primaries, a D65 white point, and the sRGB transfer function.

ValZapod:
>For my education, could you share sample files with different settings

Allmost all hdr files (maybe exceot for Matrix and Muncil UHD benchmark disk masteted on BT.2020 laser). This file from open source Netlfix archive. 
http://download.opencontent.netflix.com.s3.amazonaws.com/SolLevante/imf/SolLevante_IMF_DolbyVision_PQP3D65_UHD_24fps/VIDEO_e4da5fcd-5ffc-4713-bcdd-95ea579d790b.mxf

Many: https://trac.ffmpeg.org/search?q=%22Display+p3%22

PR #1644, opened 2022-12-28
Structs with the same name in the same namespace confuses GCC

g-maxime:
Fixes MediaArea/MediaInfo#666


JeromeMartinez:
> confuses GCC

Ha, reason I was not able to reproduce with MSVC... Thank you for the fix!
And I need to refactor/merge some code, for sure...

PR #1670, opened 2023-02-18
remove  && CPN1<'9'

pavel-pimenov:


JeromeMartinez:
Thanks for catching the issue, but the fix is wrong, `CPN1` should be replaced by `CPN2` rather than removed.

JeromeMartinez:
@pavel-pimenov mistake? the update is right but branch is deleted so I can not merge it.

Issue #1414, opened 2021-07-22
Id3v2 -- should recognize "Blues" as a valid genre


mitchblank:
I was writing some small scripts to analyze the mp3s I have in my OS/X Music library, looking for metadata inconsistencies.  To do that I have some python code that makes calls out to `mediainfo --Output=JSON -f` on each file and then turning them into python objects.  All of that is working great.

The only odd bug that I couldn't figure out is that I had a couple ripped CDs with "Genre" == "Blues" and that field was not showing up correctly.  Indeed, Music.app (f/k/a iTunes) showed the genre just fine but `mediainfo`'s CLI reported that the file had not been tagged with a genre!

That led me down a bit of a rabbit hole, but I think I understand the problem.  As you probably remember, ID3 can tag genres either as free-form text or as numeric codes.  Apple Music will support either, but if you use a genre that has a well-known ID3v1 numeric code it will use that (likely for maximum compatibly with ancient mp3-reading software)

The numeric code for "Blues" happens to be zero (see https://en.wikipedia.org/wiki/List_of_ID3v1_Genres )  MediaInfoLib knows this as well (see `Genre_000` in Source/Resource/Text/Language/DefaultLanguage.csv )

However, the code that interprets the numeric genre codes treats both `"0"` and `"255"` as unset values.  In Source/MediaInfo/Tag/File_Id3v2.cpp :
```
        case Elements::TCON :
[...]
                                if (Element_Value==__T("0") || Element_Value==__T("255"))
                                    Element_Value.clear();
                                Fill(Stream_General, 0, General_Genre, Element_Value);
[...]
        case Elements::TCO  :
[...]
                                if (Element_Value==__T("0") || Element_Value==__T("255"))
                                    Element_Value.clear();
                                if (!Element_Value.empty())
                                    Fill(Stream_General, 0, General_Genre, Element_Value);
```
So both in the code to handle `TCO` and `TCON` tags ("Content type") will treat a string with the value `"0"` as if it were unset.  This means any file that Music.app sets to have genre "Blues" is considered genre-less by MediaInfoLib.

I assume that this may have been done as a workaround for buggy mp3 generators that may have left 0 or 255 as incorrect ways of indicating a missing genre.  However, I can't find any documentation about that.  I checked in the git history and the logic seems unchanged since the code was first imported into a VCS, and there isn't anything related to "blues" in the changelog.

[...]

PR #1708, opened 2023-03-07
ID3v2: fix Genre not showing ID 0 (Blues)

JeromeMartinez:
0 is actually is valid value, only 255 is not a valid value.

Fix https://github.com/MediaArea/MediaInfo/issues/682.


g-maxime:
Also fixes #1414

Issue #1709, opened 2023-03-10
FLAC has strange sample count in Advanced mode


ValZapod:
[dafa.zip](https://github.com/MediaArea/MediaInfoLib/files/10946667/dafa.zip)

So this file has in STREAMINFO 

Samples: 1875 (0x000000753) - (36 bits)

yet in Advanced mode:

![image](https://user-images.githubusercontent.com/31514790/224440354-23ec51c4-ecc3-4220-bc30-fb9ab0676c0e.png)

File created with ffmpeg but after reencoded with ./flac same happens.

JeromeMartinez:
We have some precision issues, right... This one is easy to fix so fixed, at long term we need to handle intermediate storage differently, planned but no ETA.

PR #1710, opened 2023-03-11
FLAC: fix sampling count

JeromeMartinez:
Fix https://github.com/MediaArea/MediaInfoLib/issues/1709


ValZapod:
So now you directly read Audio_SamplingCount from STREAMINFO even though that may be fake number?

JeromeMartinez:
> So now you directly read Audio_SamplingCount

It was already the case, just that rounded duration was computed then stored (ms precision) then sampling count was computed later from the non-precise duration.

> even though that may be fake number?

MediaInfo first purpose is not to avoid to trust header, often it relies on such kind of metadata, nothing new.
Sometimes we do so coherency checks, often not (mainly because it is more work and sponsors are not interested in all formats).

PR #1726, opened 2023-04-01
USAC: fix usage of NULL in memset

nekopsykose:
NULL in c++ is not defined to always be 0, it can also be std::nullptr_t.

it is not valid to pass std::nullptr_t to memset, so pass 0 explicitly.


JeromeMartinez:
Oops, what an ugly typo, it should be an `int` for sure. Thanks.

PR #1731, opened 2023-04-06
Fix for UMR

towel42-com:
There is a UMR that throws assertions in visual studio


JeromeMartinez:
We missed this one, thanks for the fix.

PR #1753, opened 2023-04-25
Support ST381-4

ammonriley:
This patch adds support for recognizing AAC wrapped in MXF files following the ST 381-4:2017 standard.  This format is used by Harmonic's Spectrum product line to generate proxies during record.  There are a few other vendors in the broadcast industry that also generate files following this standard, but I don't have an exhaustive list.  For a sample file, see [https://harmonicinc.box.com/v/st381-4-sample]().  I don't have sample files to cover all potential variants.

Note that ST 381-4 also includes details about an AacSubDescriptor, but there doesn't seem to be any (immediate) value in parsing it.  I'd be happy to add it, if necessary.

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mxf.cpp
+++ b/Source/MediaInfo/Multiple/File_Mxf.cpp
@@ -803,6 +803,9 @@ static const char* Mxf_EssenceContainer(const int128u EssenceContainer)
                                                                                         case 0x10 : return "AVC";
                                                                                         case 0x11 : return "VC-3";
                                                                                         case 0x13 : return "Timed Text";
+                                                                                        case 0x16 : return "AAC (ADIF)";
+                                                                                        case 0x17 : return "AAC (ADTS)";
+                                                                                        case 0x18 : return "AAC (LATM/LOAS)";
                                                                                         case 0x1C : return "ProRes";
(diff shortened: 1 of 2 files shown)


ammonriley:
> ChooseParser__Aaf_GC_Sound() should not check the essence UL
Ah, okay.  I was sort of following my nose on that, and figured I'd add it for completeness.  Good to know it's not really needed.

> I note that the MXF sampling rate is 48000 despite the AAC stream is 44100, I see nothing in ST 381-4 saying that it should be forced to 48000 so I let it considered as a container/stream mismatch.
That's an error on my part.  I didn't want to go through legal to okay distribution of content recorded from our house feed, so created the sample file via an emulator, using CC0 material.  The emulator doesn't have a way to specify a different sampling rate, because (IIRC) Spectrum always records audio at 48kHz (so just fills that value in automatically), and it didn't cross my mind to resample the audio.

JeromeMartinez:
> Ah, okay. I was sort of following my nose on that,

You were courageous, because it is not a part of the code I am very proud about 😅.
The MXF parser definitely need a refactoring and documentation...

> The emulator doesn't have a way to specify a different sampling rate

No problem, I just let it as is as the mismatch catch is legitimate.

PR #1814, opened 2023-06-27
DTS-UHD: Adding parsing of .dtsx files.

RoyFunderburk:
Created a new file parser File_DtsUhd to parse DTS:X P2 input files.  Functions specific to DTSUHD parsing used by File_MPeg4_Elements.cpp are moved into File_DtsUhd.cpp, similar to how File_Dts.cpp is arranged.  This pull request implements parsing of DTS-UHD frames to provide audio information.  The frame field display provided by "--Display=1" is not implemented at this time.  I would first like to confirm the code in this pull request is OK, sort out any changes necessary, and that the "--Display=1" functionality is needed.

### DIFF
--- a/Project/BCB/Library/MediaInfoLib.cbproj
+++ b/Project/BCB/Library/MediaInfoLib.cbproj
@@ -215,6 +215,9 @@
         <CppCompile Include="..\..\..\Source\MediaInfo\Audio\File_Dts.cpp">
             <BuildOrder>89</BuildOrder>
         </CppCompile>
+        <CppCompile Include="..\..\..\Source\MediaInfo\Audio\File_DtsUhd.cpp">
+            <BuildOrder>239</BuildOrder>
+        </CppCompile>
         <CppCompile Include="..\..\..\Source\MediaInfo\Audio\File_ExtendedModule.cpp">
             <BuildOrder>99</BuildOrder>
         </CppCompile>
--- a/Project/CMake/CMakeLists.txt
+++ b/Project/CMake/CMakeLists.txt
@@ -174,6 +174,7 @@ set(MediaInfoLib_SRCS
   ${MediaInfoLib_SOURCES_PATH}/MediaInfo/Audio/File_Dsdiff.cpp
   ${MediaInfoLib_SOURCES_PATH}/MediaInfo/Audio/File_Dsf.cpp
   ${MediaInfoLib_SOURCES_PATH}/MediaInfo/Audio/File_Dts.cpp
+  ${MediaInfoLib_SOURCES_PATH}/MediaInfo/Audio/File_DtsUhd.cpp
   ${MediaInfoLib_SOURCES_PATH}/MediaInfo/Audio/File_DolbyAudioMetadata.cpp
   ${MediaInfoLib_SOURCES_PATH}/MediaInfo/Audio/File_DolbyE.cpp
   ${MediaInfoLib_SOURCES_PATH}/MediaInfo/Audio/File_ExtendedModule.cpp
--- a/Project/CodeBlocks/Library/MediaInfoLib.cbp
+++ b/Project/CodeBlocks/Library/MediaInfoLib.cbp
@@ -60,6 +60,8 @@
 		<Unit filename="..\..\..\Source\MediaInfo\Audio\File_Au.h" />
 		<Unit filename="..\..\..\Source\MediaInfo\Audio\File_Dts.cpp" />
 		<Unit filename="..\..\..\Source\MediaInfo\Audio\File_Dts.h" />
+		<Unit filename="..\..\..\Source\MediaInfo\Audio\File_DtsUhd.cpp" />
+		<Unit filename="..\..\..\Source\MediaInfo\Audio\File_DtsUhd.h" />
 		<Unit filename="..\..\..\Source\MediaInfo\Audio\File_Flac.cpp" />
 		<Unit filename="..\..\..\Source\MediaInfo\Audio\File_Flac.h" />
 		<Unit filename="..\..\..\Source\MediaInfo\Audio\File_Latm.cpp" />
--- a/Project/GNU/Library/Makefile.am
+++ b/Project/GNU/Library/Makefile.am
@@ -60,6 +60,7 @@ lib@MediaInfoLib_LibName@_la_SOURCES = \
                        ../../../Source/MediaInfo/Audio/File_Dsdiff.cpp \
                        ../../../Source/MediaInfo/Audio/File_Dsf.cpp \
                        ../../../Source/MediaInfo/Audio/File_Dts.cpp \
+                       ../../../Source/MediaInfo/Audio/File_DtsUhd.cpp \
                        ../../../Source/MediaInfo/Audio/File_DolbyAudioMetadata.cpp \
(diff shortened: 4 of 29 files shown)


JeromeMartinez:
You were courageous to adapt all the config files, and it seems perfect!

I added a quick and dirty commit with:
- parsing also from MP4 (seems that you worked with a raw stream; please provide one with DTSHDHDR so I can check this part of the code), it does not provide additional info (maybe later?) but the MediaTrace part is doable.
- Adapted `Synched_Test`, it must accept DTSUHD_NONSYNCWORD and fill FrameSize, fixed the algo issue with DTSUHD_NONSYNCWORD and CRC computing
- Adapted `Data_Parse()` for parsing several frames in practice.
- replaced some names by the spec ones (without "un", "uc" etc), not finished, please adapt the code for using the names from the spec
- Started the MediaTrace part, was easier for me to understand the code, but not finished (and all theses lines will be changed when the bitstream parser is replaced)

Could you provide a file with MD01 buffer? With that, I would check and replace the bitstream parser you use in order to use the MediaInfo one.

Specs stipulate "When m_bFullChannelBasedMixFlag = FALSE, the FTOC CRC16 word is transmitted in every frame, so the CRC
check is performed in all frames." but it seems that it is when m_bFullChannelBasedMixFlag is TRUE with your files, could you confirm that this is expected?

RoyFunderburk:
I can confirm FullChannelMixFlag as false indicates CRC present in non-sync frames.  This check is correct in ParseStreamParams, but I have a bug in File_DtsUhd::CheckCurrentFrame where it is using the  Buffer+Buffer_Offset instead of FrameStart+FrameSize.  Since Buffer+Buffer_Offset is the sync frame, the CRC was checked on the sync frame instead of the non-sync frame as it should have been. I will update the code to correct that.

Currently our encoder does not generate the MD01 data, meaning File_DtsUhd::ReadBitsMD01(MD01* MD01, int Bits)  will always have MD01->Buffer.empty() as true.  Would adding a comment to the code describing this situation be OK as a resolution for this?

I will add to your patch the changes for the names to match the specification, as well as the correction for the CRC check.

Thanks,
-Roy

JeromeMartinez:
ReadBits completely removed in last commit so we are sure that all is handled by MediaInfo bitstream parser.

RoyFunderburk:
Only found a few minor things that needed changing, the bit reading adaptation looks fine.

JeromeMartinez:
Ugly typo on my side about `ChLayoutIndex` but having a file on my side with that would help to avoid such typo ;-).

I disagree with DTSHDHDR handling, there is a mix of DTSHDHDR data and UHD frame data in the computing of some elements, but DTSHDHDR has everything (standalone).

As DTSHDHDR is also for DTS or DTS-HD (not UHD), I implemented a full and standalone DTSHDHDR parser also used by DTS parser and rebased.

About sample count, DTSHDHDR has the count before frame encoding, it is better to use this one, I implemented using `Num_Samples_Orig_Audio_At_Max_Fs` (for information that leads to a slightly different duration between the raw stream and the MP4 because the MP4 `stss` does not store the real sample count for the last frame but use the sample count of the total frame so with padding samples).

`TC_Frame_Rate` is wrongly used as DTS-UHD frame rate but they are different, I adapted the DTSHDHDR  parser in order to use it only for time code info (which is now displayed, by the way). DTS-UHD frame rate is now automatically computed from sample rate and samples per frame.

DTS-UHD samples per frame seems wrong (not the same as in DTSHDHDR), I adapted the code and now it seems OK.

`MaxPayloadCode` seems to provide the maximum theoretical bitrate, but DTSHDHDR provides the real one, I opted to use `MaxPayloadCode` only if DTSHDHDR does not provide the real maximum bitrate (on our side maximum bitrate is expected to be the real one if available else the theoretical one as fallback).
Note: maximum bitrate is not yet computed from MP4 header but planned, so it will be as with the raw stream.

Removed (again? I think it the issue was already in previous PR and I removed theses lines) the `if (RepresentationType==3) //Binarual is always two channels`, I don't see that in spec and ch3.mp4 becomes 1-channel with that, which seems wrong to me (only left channel?), additionally it is not coherent with ch3.dtsx, please more details about such change if you want it.

Some cleanup. The `--Details` output is nice now :).

Please review changes.

PS: there would be a lot to peek from DTSHDHDR and crosscheck with DTS and DTS-UHD frames, but not the purpose of the task, maybe one day if there is a sponsor interest in that.

RoyFunderburk:
I was a bit confused on the frame rate but I've got a better understanding of the frame rate now.  The Reptype==3 is fine the way you changed it, I had mistakenly used the activity mask of 2 instead of the channel mask which was 6. The max payload code from the DTSHDHDR chunks is better, since it cannot be determined from a single audio frame.

I did not find any issues with the changes you made.

The details output looks great, thank you for putting the time into this.

JeromeMartinez:
> I had mistakenly used the activity mask of 2 instead of the channel mask which was 6.

Even if it was 6, I think I prefer to show the stored value rather than forcing a value, so I keep the lines removed.

> I did not find any issues with the changes you made.

I rebased and updated the code with the parsing of the buffered content, for tests I cheated with p2.mp4 and forced the fill of the buffer, demo is in last commit (in comment), so I removed ReadBitsMD01, please review the new commit, and if you have non shareable files with buffers please test with this code.

I think it is the last update from me, if fine for you I'll squash the commits and merge.

RoyFunderburk:
The latest changes look good to me, please go ahead and merge.  Thanks!

RoyFunderburk:
I added reporting of LKFS value, and the IMAX format, if this causes any issues I can revert the commit.   A DTS:X P2 file (imax.dtsx) is available in the media-samples repository.

JeromeMartinez:
>  and the IMAX format

IIUC this is just a specific channel config, even not an extra feature, so I'll move IMAX to "commercial name" when such config is seen (because it is just a DTS-UHD file without any extra requirement so DTS-UHD format), beside a couple of adaptations (quicker to update myself than requesting changes :-p).

JeromeMartinez:
Small more or less technical update.

But a classic 5.1 stream is "IMAX" just because ChActivityMask is used for something doable with ChLayoutIndex (so useless) instead of using ChLayoutIndex directly, really???
Any public documentation about that?

RoyFunderburk:
Thanks for accepting this update so quickly.  The addition of IMAX to P2 was a more recent change and the ETSI spec is not yet updated for this.  I don't know the details on why the ChActivityMask was used to indicate IMAX, but yes, that the the method of indicating IMAX in UHD.

JeromeMartinez:
> I don't know the details on why the ChActivityMask was used to indicate IMAX, but yes, that is the method of indicating IMAX in UHD.

What is the behavior of a decoder supporting ETSI TS 103 491 V1.2.1 only (so no IMAX stuff) compared to a decoder supporting the next version of ETSI TS 103 491 with IMAX stuff?

JeromeMartinez:
@RoyFunderburk I squashed your 2 last commits.

RoyFunderburk (commented):
I approve of these changes.  Github will not let me select the "Approve" option (authors cannot approve their own pull request).

RoyFunderburk (commented):
Also I approve of the changes to the copyright, especially the "Provided by xperi".

PR #1830, opened 2023-07-11
Move CURL info message to stderr

g-maxime:


JeromeMartinez:
It seems that it was some debug stuff :(.
@g-maxime please keep this change but also comment out the whole "else" block, we actually don't need it in production.

PR #1832, opened 2023-07-12
DTS Core IMAX detection

RoyFunderburk:
.dtshd files already have detection for IMAX in the extension frames, this pull request adds IMAX detection for core-only .dtshd files.  There is a flag indicating IMAX content in the Rev2 Auxiliary data chunk, parsing for this chunk is added.  In order to parse this chunk, a field from further down in the core header (sub sub frame count) is needed, so the core chunk processing is expanded a bit also.  I tried to keep the addition consistent with the existing IMAX detection.

New sample files "rev2aux-X.dts" test variations on the Rev2Aux chunk.  imax.dtshd is a core-only file with the rev2aux chunk indicating IMAX content.  61.dtshd and 121.dtshd have IMAX in extension frames that test the pre-existing IMAX detection.  in.dtshd is similar to imax.dtshd, but is normal, not IMAX.


Source/MediaInfo/Audio/File_Dts.cpp
```diff
@@ -2297,6 +2369,80 @@ void File_Dts::XBR()
...
+            Get_SB(DRCMetadataPresent,                          "DRCMetadataPresent");
+            Get_SB(DialnormMetadataPresent,                     "DialnormMetadataPresent");
+            if (DRCMetadataPresent)
+                Skip_S1(4,                                      "DRCversion_Rev2AUX");
+            auto Remain=Data_BS_Remain()%8;
+            if (Remain)
+                Skip_S1(Remain,                                 "ByteAlign");
+            if (DRCMetadataPresent)
+                for (int i=0; i<SubSubFrameCount; i++)
+                    Skip_S1(8,                                  "subsubFrameDRC_Rev2AUX");
```
JeromeMartinez, line 2408: "DRCCoeff_Rev2"

Source/MediaInfo/Audio/File_Dts.cpp
```diff
@@ -2297,6 +2369,80 @@ void File_Dts::XBR()
...
+void File_Dts::Rev2Aux()
+{
+    // ETSI TS 102 114 v1.6.1 Section 5.8.3.
+    // The Rev2 Auxiliary Data Chunk Structure contains the T1CC indicator
+    // in the reserved data.
+    int8u Rev2Bytes;
+    Peek_B1(Rev2Bytes);
+    bool DmixIndexPresent=(Rev2Bytes&1)==1;
+    Rev2Bytes = (Rev2Bytes>>1)+1;
+    if (Dts_CRC_CCIT_Compute(Buffer+Buffer_Offset+Element_Offset, Rev2Bytes)==0)
```
JeromeMartinez, line 2382: Add `Element_Size-Element_Offset>=Rev2Bytes &&` check.

JeromeMartinez (changes requested):
Generally speaking, I don't get the meaning of the `Type1CertifiedContent`, it seems that it is only a flag without actual new content (only few bytes of metadata?) so just some marketing stuff, what is the technical difference? In other words, will a decoder not supporting this extension decode differently the stream compared to a decoder supporting this extension, and will a decoder supporting this extension decode differently a stream with this extension than a stream without this extension (e.g. if I replace 0x7004C070 by 0x12345678, is the decoded content same between the modified file and the unmodified file?)

With MediaInfo we do care about the difference between marketing and technical parts, so this does matter to know if a "legacy" decoder can decode in the same manner or not such stream.

Source/MediaInfo/Audio/File_Dts.cpp
```diff
@@ -2297,6 +2369,80 @@ void File_Dts::XBR()
...
+        bool BroadcastMetadataPresent = false;
+        Get_SB(BroadcastMetadataPresent,                        "BroadcastMetadataPresent");
+        if (BroadcastMetadataPresent)
+        {
+            bool DRCMetadataPresent, DialnormMetadataPresent;
+            Get_SB(DRCMetadataPresent,                          "DRCMetadataPresent");
+            Get_SB(DialnormMetadataPresent,                     "DialnormMetadataPresent");
+            if (DRCMetadataPresent)
+                Skip_S1(4,                                      "DRCversion_Rev2AUX");
+            auto Remain=Data_BS_Remain()%8;
```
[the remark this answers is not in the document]
RoyFunderburk: With the "Rev2_DRCs", the specification has the "DRCversion_Rev2AUX" field labeled as "DRCversion_Rev2AUX", so I am guessing you might have meant that would be a good title for an Element around the "subsubFrameDRC_Rev2AUX" array.  Also I changed "subsubFrameDRC_Rev2AUX" to "DRCCoeff_Rev2 array".

RoyFunderburk:
> if I replace 0x7004C070 by 0x12345678, is the decoded content same between the modified file and the unmodified file?)

All of the channels decode exactly the same data for these .dtshd imax files.  The difference is the Ls/Rs channels are used as Lsr/Rsr.  I had missed that detail in the pull request changes and I will update the channel display to reflect that situation.

JeromeMartinez:
>  The difference is the Ls/Rs channels are used as Lsr/Rsr

For 5.1? It seems a very small change, because in practice it does not change the user experience, Ls/rs and Lsr/Rsr being very similar for 5.1 config, and maybe same in practice.
In practice, with 6 speakers at home, what a decoder aware of Type1CertifiedContent will do a decoder not aware of Type1CertifiedContent won't do?

the format profiles part is intended for differences if a decoder does not support a feature, e.g. a 7.1 content is played as 5.1 (2 channels completely ignored) or a lossless content is not decoded as lossless, here a flag and no "visible" difference in practice makes me think that I'll move imax to the "marketing" part only.

JeromeMartinez (changes requested):
Generally speaking, I don't get the meaning of the `Type1CertifiedContent`, it seems that it is only a flag without actual new content (only few bytes of metadata?) so just some marketing stuff, what is the technical difference? In other words, will a decoder not supporting this extension decode differently the stream compared to a decoder supporting this extension, and will a decoder supporting this extension decode differently a stream with this extension than a stream without this extension (e.g. if I replace 0x7004C070 by 0x12345678, is the decoded content same between the modified file and the unmodified file?)

With MediaInfo we do care about the difference between marketing and technical parts, so this does matter to know if a "legacy" decoder can decode in the same manner or not such stream.

RoyFunderburk:
> In practice, with 6 speakers at home, what a decoder aware of Type1CertifiedContent will do a decoder not aware of Type1CertifiedContent won't do?
> it seems that it is only a flag without actual new content (only few bytes of metadata?)

The .dtshd decoder will output the same PCM data regardless of the t1cc setting, the intention was the audio that is used for the encode would be different.

RoyFunderburk:
I modified the t1cc/imax detection to only report t1cc/imax in the "Commercial name" as you suggested.  There are no changes to the format or any other report fields.

JeromeMartinez:
I tweaked a bit the output for doing the difference between the technical flag and the commercial name.

JeromeMartinez (approved):

JeromeMartinez (changes requested):
Mismatch between specs and provided code/files

PR #1836, opened 2023-07-13
DTS-UHD: Report t1cc flag in commercial name

RoyFunderburk:
This pull request adds the T1CC flag to DTS-UHD in the commercial name only.


JeromeMartinez:
@RoyFunderburk apologies, we were not able to review the last PRs before the release and we need to release, so DTS-UHD support will be in next release but without t1cc flag support. We'll handle that for the next release (it won't be long).

JeromeMartinez:
I tweaked a bit the output for doing the difference between the technical flag and the commercial name.

JeromeMartinez (approved):

PR #1922, opened 2023-12-01
GXF: support of AVC and VC-3

JeromeMartinez:
Fix [SF537](https://sourceforge.net/p/mediainfo/feature-requests/537/).


Source/MediaInfo/File__Analyze_Buffer.cpp
```diff
@@ -2049,7 +2049,7 @@ void File__Analyze::Skip_XX(int64u Bytes, const char* Name)
 {
     if (Element_Offset+Bytes!=Element_TotalSize_Get()) //Exception for seek to end of the element
     {
-        INTEGRITY_SIZE_ATLEAST(Bytes);
+        //INTEGRITY_SIZE_ATLEAST(Bytes);
```
cjee21, line 2052: @JeromeMartinez why was integrity check disabled by this commit/PR? Also the INTEGRITY_SIZE_ATLEAST macro might be vulnerable to overflow if Bytes is very large.

cjee21: @JeromeMartinez Have you seen the previous message above?

cjee21: Integrity checks restored and improved in https://github.com/MediaArea/MediaInfoLib/pull/2570

PR #1946, opened 2024-01-18
tfsxml: fix sign integer type

robUx4:
get_level() expects an int pointer.


JeromeMartinez (changes requested):
This creates a signed/unsigned mismatch in other places
I prefer the other way (unsigned everywhere).

JeromeMartinez (approved):

PR #1948, opened 2024-01-18
fix build with MEDIAINFO_DVDIF_ANALYZE_YES is not set

robUx4:


JeromeMartinez (changes requested):
Does not compile everywhere.

Replace by
~~~
    #if defined(MEDIAINFO_DVDIF_ANALYZE_YES)
    #endif //defined(MEDIAINFO_DVDIF_ANALYZE_YES)
~~~

JeromeMartinez (approved):

PR #1969, opened 2024-02-24
Add Atmos channel layouts from CoreAudioTypes.h.

gmcgarry:
As seen at the following locations:

- https://developer.apple.com/documentation/coreaudiotypes/1572103-audio_channel_labels
- https://github.com/xybp888/iOS-SDKs/blob/master/iPhoneOS13.0.sdk/System/Library/Frameworks/CoreAudioTypes.framework/Headers/CoreAudioBaseTypes.h

Before:

```
Audio
Format                                   : ALAC
Codec ID                                 : alac
Codec ID/Info                            : Apple Lossless Audio Codec
Duration                                 : 28 s 96 ms
Bit rate                                 : 349 kb/s
Channel(s)                               : 12 channels
Channel layout                           : L R C LFE Lb Rb Ls Rs 46 48 52 54
Sampling rate                            : 48.0 kHz
Stream size                              : 1.17 MiB (98%)

```

After:

```
Audio
Format                                   : ALAC
Codec ID                                 : alac
Codec ID/Info                            : Apple Lossless Audio Codec
Duration                                 : 28 s 96 ms
Bit rate                                 : 349 kb/s
Channel(s)                               : 12 channels
Channel layout                           : L R C LFE Lb Rb Ls Rs Ltf Rtf Ltr Rtr
Sampling rate                            : 48.0 kHz
Stream size                              : 1.17 MiB (98%)

```

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mpeg4_Elements.cpp
+++ b/Source/MediaInfo/Multiple/File_Mpeg4_Elements.cpp
@@ -467,6 +467,16 @@ static string Mpeg4_chan_ChannelDescription_Layout (int32u ChannelLabel)
         case  43 : return "DialogCentricMix";
         case  44 : return "CenterSurroundDirect";
         case  45 : return "Haptic";
+        case  46 : return "Ltf";
+        case  47 : return "Ctf";
+        case  48 : return "Rtf";
+        case  49 : return "Ltm";
+        case  50 : return "Ctm";
+        case  51 : return "Rtm";
+        case  52 : return "Ltr";
+        case  53 : return "Ctr";
+        case  54 : return "Rtr";
+
         case 200 : return "W";
         case 201 : return "X";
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
Yet other acronyms...

there is a need to update [mapping from format specific to MediaInfo names](https://github.com/MediaArea/MediaInfoLib/blob/master/Source/MediaInfo/MediaInfo_Internal.cpp#L383-L386) ([MediaInfo Names](https://mediaarea.net/AudioChannelLayout)).
But it won't work out of the box due to [that](https://github.com/MediaArea/MediaInfoLib/blob/master/Source/MediaInfo/MediaInfo_Internal.cpp#L396) (need to add ALAC there, or something about MOV).

Please a sample file.
Would be nice if you do the mapping from Apple names to MediaInfo names, I may manage the other issues if I have a sample file.

Issue #2021, opened 2024-04-22
Number of channels in not returned as Channels(s) property in last version


TondaKrist:
Files recorded on BlackMagic Hyperdeck in H264, does not display at all` Channel count` property in audio streams.
Works fine in OLD version of MediaInfo (0.7) and returns its value in `Channel(s)` property properly.

When Debug mode (Details - 5) is set, then it shows that some `channelcount` value is present, but not returned in properties.

File to reproduce
https://github.com/MediaArea/MediaInfo/assets/17853744/31e0f524-8173-4ce4-a681-19dd7d6b8346

JeromeMartinez:
The "2" in the previous versions was not right for files with something else than 2 channels because the "2" is hardcoded in this place for MP4 (it is from QuickTime) so we removed the corresponding code for non QuickTime files.

Then it seems that the file is not conform to MP4, something like if it is a QuickTime file but signaling ISO MP4, so the lack of detection of the format (AAC) and its parsing.
We see that FFmpeg succeeds to decode it, we'll check what is their algorithm for differentiating ISO MP4 style from QucikTime style without relying on `ftyp`.

JeromeMartinez:
The file is confirmed to be buggy but we use now the same trick as FFmpeg for catching Qt style AudioSampleEntry in ISO MP4.

PR #2022, opened 2024-04-23
MP4: support of Qt style AudioSampleEntry in ISO MP4

JeromeMartinez:
Some files may have Qt style AudioSampleEntry in ISO MP4 When AudioSampleEntry is v1, it may be easily detected by stsd version 0 (which is forbidden by ISO MP4)

Fix https://github.com/MediaArea/MediaInfoLib/issues/2021.


TondaKrist:
Thanks a lot.
I will report to BlackMagic that files recorded on their recorder is buggy :)

PR #2071, opened 2024-06-12
WavPack: various fixes for multichannel & DSD files

dbry:
- fix parsing of multichannel files with greater than 8 streams
- fix compression ratio calculation for all multichannel files
- fix DSD file parsing

### DIFF
--- a/Source/MediaInfo/Audio/File_Wvpk.cpp
+++ b/Source/MediaInfo/Audio/File_Wvpk.cpp
@@ -154,7 +154,7 @@ void File_Wvpk::Streams_Finish()
         int64u BitDepth=dsf?1:(Wvpk_Resolution[(resolution1?1:0)*2+(resolution0?1:0)]);
         int64u Duration=Samples*1000/SamplingRate;
         int64u CompressedSize=File_Size-TagsSize;
-        int64u UncompressedSize=Duration*(mono?1:2)*BitDepth*(SamplingRate<<(3*dsf))/8/1000;
+        int64u UncompressedSize=Duration*(num_channels?num_channels:(mono?1:2))*BitDepth*(SamplingRate<<(3*dsf))/8/1000;
         float32 CompressionRatio=((float32)UncompressedSize)/CompressedSize;
         Fill(Stream_Audio, 0, Audio_StreamSize, CompressedSize, 3, true);
         Fill(Stream_Audio, 0, Audio_Duration, Duration, 10, true);
@@ -585,8 +585,8 @@ void File_Wvpk::Data_Parse_Fill()
     Fill(Stream_Audio, 0, Audio_Codec_Settings, Mode);
     if (dsf)
     {
-        Fill(Stream_Audio, 0, Audio_Format_Settings_Mode, "DSF");
-        Fill(Stream_Audio, 0, Audio_Format_Settings, "DSF");
+        Fill(Stream_Audio, 0, Audio_Format_Settings_Mode, "DSD");
+        Fill(Stream_Audio, 0, Audio_Format_Settings, "DSD");
     }
     if (correction)
         Fill(Stream_Audio, 0, Audio_Format_AdditionalFeatures, "Correction");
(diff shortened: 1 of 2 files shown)


dbry:
Hi!

A user reported to me that certain multichannel files were not reporting the correct number of channels in MediaInfo (e.g., 18 instead of 19) and I tracked it down and also fixed a couple other things I found.

I hope this is useful, and thanks for the ongoing WavPack support! Please let me know if there's anything else I can do.

-David

JeromeMartinez (changes requested):
> A user reported to me that certain multichannel files were not reporting the correct number of channels in MediaInfo (e.g., 18 instead of 19) and I tracked it down and also fixed a couple other things I found.

Thank you for the fixes.

> Please let me know if there's anything else I can do.

Please provide sample files for both cases (or at least one if difficult to provide both), and review the changes I forced-pushed.
Force-push from your side if you want a clean PR.

> __T("DSD")+Ztring::ToZtring(16<<SamplingRate_Shift))

I am fine with the `DSD` change ([DSD](https://en.wikipedia.org/wiki/Direct_Stream_Digital) is the content format, DSF a file format), but the mode shall not include the sampling rate which is a parameter and already elsewhere (in the "sampling rate" field).

>  Get_L4 (channel_mask,                               "channel_mask");

Now there is code for more than 256 channels but channel mask accepts only 32 cases, how is it working for mapping 33 channels to their positions?

JeromeMartinez:
Additional question:

> fix parsing of multichannel files with greater than 8 streams

`0D` with `Size` < 6 was supporting up to 255 channels, why do you speak about 8 and why do you speak about streams (from the new code, "stream" seems to refer to something else)? Or does it means that in the past a "single "stream" was not supporting more than 8 channels so 9 channels requires this update for working well?

dbry:
> > A user reported to me that certain multichannel files were not reporting the correct number of channels in MediaInfo (e.g., 18 instead of 19) and I tracked it down and also fixed a couple other things I found.
> 
> Thank you for the fixes.
> 
> > Please let me know if there's anything else I can do.
> 
> Please provide sample files for both cases (or at least one if difficult to provide both), and review the changes I forced-pushed. Force-push from your side if you want a clean PR.

[384 channels](https://www.wavpack.com/384-channel-sample.wv)

[18 channels](https://www.wavpack.com/18-channel-sample.wv)

The [WavPack test suite](http://www.rarewares.org/wavpack/test_suite.zip) has a DSD file (bit_depths/1bit_dsd.wv)

I have reviewed and tested your changes. No issues found. Thanks!

> 
> > __T("DSD")+Ztring::ToZtring(16<<SamplingRate_Shift))
> 
> I am fine with the `DSD` change ([DSD](https://en.wikipedia.org/wiki/Direct_Stream_Digital) is the content format, DSF a file format), but the mode shall not include the sampling rate which is a parameter and already elsewhere (in the "sampling rate" field).

I know that the sampling rate field provides that information, but nobody knows what those values mean. Everyone knows what `DSD64` or `DSD128` mean and I would say that those _are_ the format's names in the popular lexicon. But it's your call.

> 
> > Get_L4 (channel_mask,                               "channel_mask");
> 
> Now there is code for more than 256 channels but channel mask accepts only 32 cases, how is it working for mapping 33 channels to their positions?

Any channels that exist that are not represented by a set bit in the channel mask are simply "undefined" speakers (the defined ones must come first). There is a mechanism to define these channels that I implemented to handle CAF channel definitions, but nobody is using CAF so it's all rather moot (if you're curious you can look up `WavpackGetChannelIdentities()` in the [libwavpack documentation](https://www.wavpack.com/WavPack5LibraryDoc.pdf)). The applications that use large number of channels (e.g., high-order Ambisonics or [electrophysiology data](https://www.biorxiv.org/content/10.1101/2023.05.22.541700v1.full.pdf)) know what the channels mean.

dbry:
> Additional question:
> 
> > fix parsing of multichannel files with greater than 8 streams
> 
> `0D` with `Size` < 6 was supporting up to 255 channels, why do you speak about 8 and why do you speak about streams (from the new code, "stream" seems to refer to something else)? Or does it means that in the past a "single "stream" was not supporting more than 8 channels so 9 channels requires this update for working well?

Multichannel files are split into a sequence of mono and stereo "streams". For example, 5.1 would normally be represented with a stereo stream for FL+FR, a mono stream for FC, a mono stream for LFE, and a stereo stream for BL+BR. Frames containing audio data would cycle through those 4 streams (with the first marked `INITIAL_BLOCK` and the last flagged `FINAL_BLOCK`).

Before version 5.0, WavPack files could only have a _maximum of 8 such streams_, which results in a maximum of 8 channels (assuming no pairing) to 16 channels (assuming all channels paired into stereo blocks).

For version 5.0 I wanted to eliminate that limit and so I created the new version of the `ID_CHANNEL_INFO` block with 6 or 7 bytes. It specifies _both_ the number of channels and the number of streams (so they can be pre-allocated), and allows up to 4096 channels and streams. To reduce breakage of old decoders (like FFmpeg) I only use this variation if there are more than 8 streams. Old decoders will error on the metadata block being 6+ bytes, but most multichannel files will still work fine.

JeromeMartinez:
> I know that the sampling rate field provides that information, but nobody knows what those values mean. Everyone knows what DSD64 or DSD128 mean and I would say that those are the format's names in the popular lexicon. 

I'll think to something generic (about DSD sampling rate, not WavPack) about that.

> The applications that use large number of channels (e.g., high-order Ambisonics or [electrophysiology data](https://www.biorxiv.org/content/10.1101/2023.05.22.541700v1.full.pdf)) know what the channels mean.

OK, so for the moment no support of something else than the current 32-bit channel mask.

PR #2084, opened 2024-06-26
HTML Improvements 2

cjee21:
More HTML improvements:
- Enable native dark mode
  - There is now dark mode for the exported HTML on all modern browsers
  - Will work in MediaInfo GUI too with newer browser engine implemented
- Move all styles from inline to <style> tag
- Remove colspans that appear to do nothing and cause validator errors
- Standardize indents

HTML validator results:
Screenshot 2024-06-28 at 17-01-08 Showing results for uploaded file Test mp4 MediaInfo html - Nu Html Checker

There is only one warning for language.

Tested for proper display and dark mode switching on Edge, Firefox and MediaInfo VCL GUI.

This change makes MediaInfo Qt GUI display differently since it is currently not using a web engine for the HTML view. This should be fixable on Qt GUI side if necessary.

### DIFF
--- a/Source/MediaInfo/MediaInfo_Inform.cpp
+++ b/Source/MediaInfo/MediaInfo_Inform.cpp
@@ -436,7 +436,23 @@ Ztring MediaInfo_Internal::Inform()
     #endif //defined(MEDIAINFO_CSV_YES)
 
     if (HTML)
-        Retour+=__T("<!DOCTYPE html>\n<html>\n<head>\n    <meta charset=\"utf-8\">\n    <meta name=\"viewport\" content=\"width=device-width, initial-scale=1\">\n    <title>MediaInfo</title>\n</head>\n<body>\n");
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
Thank you. For this PR and the MediaInfo ones related to HTML, we'll check them and merge them after the release (expected this week), in order not to change so much code at the last minute.

cjee21:
Turns out there's a CSS syntax mistake after all.  I'll correct that and update first post. Been some time since I last used `@media (prefers-color-scheme: dark)` and the example that I referred to turned out to be non-ideal.

cjee21:
Some more improvements:
- improved column sizing to hopefully prevent any chances of text being clipped
- replaced `<i>` and `<br>` with CSS
- further improved indents of generated HTML code

Tested on Edge, Firefox and MediaInfo VCL GUI (both HTML engines). HTML validator results remain the same.

PR #2092, opened 2024-07-07
HEVC: add missing parsing of some multilayer SPS elements

jamrial:
I have a few multiview samples generated by either iPhone or Apple's authoring tools, and for all of them i get the following:

```
Conformance errors                       : 1
 HEVC                                    : Yes
  General compliance                     : Bitstream parsing ran out of data to read before the end of the syntax was reached, most probably the bitstream is malformed  (offset 0xDEADBEEF)
Codec configuration box                  : hvcC+lhvC
```

In both cases, the offset reported is the start of whatever box follows the lhvC one, so it was clear there were LHEVCDecoderConfigurationRecord parsing related issues.

Looking into it i found one problem parsing SPS when nuh_layer_id > 0 and one when MultiLayerExtSpsFlag is true (nuh_layer_id > 0 && sps_ext_or_max_sub_layers_minus1 == 7), both of which i'm fixing in this PR, as the LHEVCDecoderConfigurationRecord itself seems to be ok.
Only the latter is enough to fix parsing the samples i have, but i'm including the other as well to ensure correct parsing for all files. You can check the syntax i'm fixing in section F.7.3.2.2.1 of ITU-T H.265

### DIFF
--- a/Source/MediaInfo/Video/File_Hevc.cpp
+++ b/Source/MediaInfo/Video/File_Hevc.cpp
@@ -1914,8 +1914,9 @@ void File_Hevc::seq_parameter_set()
     }
     if (nuh_layer_id)
     {
-        max_sub_layers_minus1=0;
+        const auto vps_max_sub_layers_minus1 = (*video_parameter_set_Item)->vps_max_sub_layers_minus1;
         Get_S1 (3, sps_ext_or_max_sub_layers_minus1,            "sps_ext_or_max_sub_layers_minus1");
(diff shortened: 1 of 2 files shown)


JeromeMartinez:
Thank you for the patch.

> I have a few multiview samples generated by either iPhone or Apple's authoring tools

Can you share one?
The files I received don't show this issue.

jamrial:
Yes, here: https://github.com/MediaArea/MediaInfoLib/assets/2442845/48c2a34d-1924-496f-915a-8f3db28d6ada

JeromeMartinez:
Thank you!

PR #2115, opened 2024-09-30
Escape HTML strings

### LINKED ISSUE #1975: HTML tags are not escaped in HTML output
If any fields in the media file contain HTML, they seem to be unchanged in the HTML output of MediaInfo, which might break parsing and potentially do other unwanted stuff.

Screenshot 2024-02-25 211155<br>

Fields should probably be escaped, like they are in the XML output.

cjee21:
Escape HTML strings to ensure they display exactly what is in the file and not trigger unwanted results.

Fixes #1975 

Test file:
https://github.com/user-attachments/assets/fcebf2f9-532c-4a25-b810-5a389f55d44b

Before:
Screenshot 2024-09-30 154319
Screenshot 2024-09-30 154326

After:
Screenshot 2024-09-30 154405


JeromeMartinez:
Oops!!!! 20 year old code... Thank you for catching that.

Issue #2105, opened 2024-08-09
heap-buffer-overflow in MediaInfoLib::File_Id3v2::Data_Parse() ../../../Source/MediaInfo/Tag/File_Id3v2.cpp:597


SuyueGuo:
### Summary
A heap-buffer-overflow vulnerability was found in MediaInfo, it may cause arbitrary code execution.

### Version
```
mediainfo --version
MediaInfo Command line, 
MediaInfoLib - v24.06
```

### Details

ASAN output:
```
=================================================================
==2239452==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000000232 at pc 0x7f64f24e02c3 bp 0x7fff8898ac20 sp 0x7fff8898a3c8
WRITE of size 2882 at 0x602000000232 thread T0
    #0 0x7f64f24e02c2 in __interceptor_memcpy ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:827
    MediaArea/MediaInfo#1 0x55cf77dbe957 in memcpy /usr/include/x86_64-linux-gnu/bits/string_fortified.h:29
    MediaArea/MediaInfo#2 0x55cf77dbe957 in MediaInfoLib::File_Id3v2::Data_Parse() ../../../Source/MediaInfo/Tag/File_Id3v2.cpp:597
    MediaArea/MediaInfo#3 0x55cf780500bc in MediaInfoLib::File__Analyze::Data_Manage() ../../../Source/MediaInfo/File__Analyze.cpp:2810
    MediaArea/MediaInfo#4 0x55cf7805353c in MediaInfoLib::File__Analyze::Buffer_Parse() ../../../Source/MediaInfo/File__Analyze.cpp:1941
    MediaArea/MediaInfo#5 0x55cf78053c87 in MediaInfoLib::File__Analyze::Open_Buffer_Continue_Loop() ../../../Source/MediaInfo/File__Analyze.cpp:1507
    MediaArea/MediaInfo#6 0x55cf78055767 in MediaInfoLib::File__Analyze::Open_Buffer_Continue(unsigned char const*, unsigned long) ../../../Source/MediaInfo/File__Analyze.cpp:1101
[...]

PR #2117, opened 2024-09-30
Fix out-of-bounds array read in File_Aac_Main

cjee21:
As mentioned in https://github.com/MediaArea/MediaInfoLib/issues/2105#issuecomment-2381399096, this is one of the issues found with Visual Studio Code Analysis and also Cppcheck.

Visual Studio:
Screenshot 2024-09-30 175112

Cppcheck:
```
Id: arrayIndexOutOfBoundsCond
CWE: 788
Either the condition 'OutputChannels[i]>Aac_OutputChannelPosition_Size' is redundant or the array 'Aac_ChannelMode[43]' is accessed at index 43, which is out of bounds.
```

https://github.com/MediaArea/MediaInfoLib/blob/00d96580cfaff7f189af13d6ec6cf8f7d8b3a72d/Source/MediaInfo/Audio/File_Aac_Main.cpp#L498-L501

The issue is that `Aac_OutputChannelPosition_Size` is 43 and `Aac_ChannelMode` has 43 elements. Since an array starts with `0`, the max we can access in `Aac_ChannelMode` is 42. So we change `>Aac_OutputChannelPosition_Size` to `>=Aac_OutputChannelPosition_Size` to prevent accessing `Aac_ChannelMode[43]`.

### DIFF
--- a/Source/MediaInfo/Audio/File_Aac_Main.cpp
+++ b/Source/MediaInfo/Audio/File_Aac_Main.cpp
@@ -495,7 +495,7 @@ extern string Aac_ChannelMode_GetString(const Aac_OutputChannel* const OutputCha
     memset(ChannelModes, 0, Aac_ChannelMode_Max+1);
     for (int i=0; i<OutputChannels_Size; i++)
     {
-        if (OutputChannels[i]>Aac_OutputChannelPosition_Size)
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
Ouch... Good catch.
We definitely to run (again) the static analysis... Thank you for handling that.

Issue #2120, opened 2024-10-08
Memory leak when reading compressed headers in PNGs


luis-vsn:
Hello mantainers,

I have found a memory leak when referencing the library from another project. It leaks memory when PNG files have compressed headres so they need to be uncompressed using zlib. 

PoC
[cursed.zip](https://github.com/user-attachments/files/17288523/cursed.zip)

To reproduce it, just add a loop after main analysis in HowToUse_Dll.cpp:

```cpp
MI.Close();
int i = 0;
do {
    MI.Open(__T("cursed.png"));
    MI.Close();
} while (i++ < 1000);
```

This leak does not apply when analyzing the file with CLI.

JeromeMartinez:
https://github.com/MediaArea/MediaInfoLib/pull/2121

PR #2121, opened 2024-10-08
Removed buffer leak in PNG file parsing and when using inflate

luis-vsn:
Solves recently opened issue at #2120


JeromeMartinez (approved):
Ouch, it hurts that we had such buggy code.

Source/MediaInfo/Image/File_Png.cpp
```diff
@@ -462,6 +463,7 @@ void File_Png::iCCP()
             Open_Buffer_Continue(&ICC_Parser, (const int8u*)Buffer, Buffer_Size);
             Open_Buffer_Finalize(&ICC_Parser);
             Merge(ICC_Parser, StreamKind_Last, 0, 0);
+            delete[] Buffer;
```
cjee21: @luis-vsn @JeromeMartinez `delete[] Buffer` is put inside `#if defined(MEDIAINFO_ICC_YES)` so it will not be deleted if MediaInfo is built without ICC support. Actually, why do the uncompression when ICC is not enabled? Maybe whole block should be put in `#if defined(MEDIAINFO_ICC_YES)`.

PR #2123, opened 2024-10-08
Fix version number in some project files

g-maxime:


JeromeMartinez:
missing "v" in commit message ;-).

caoccao:
I wonder if you are file with my contributing a script that can update the version of all related files as a whole so that this kind of version mismatch will no longer be a pain?

cjee21:
The current upgrade scripts are located here: https://github.com/MediaArea/MediaArea-Utils

If I understand correctly, I think the weakness of the current script is that it requires the previous version to be an exact match. So once it is out of sync, it will require manually fixing the mismatch.

I don't know how it went out of sync previously but a possibility in the future may be that someone edits something in an IDE that updates the project or resource files. IDEs will save 6 instead of 06 for example as minor version since starting with zero is technically invalid except for the strings section. 

Maybe the script can be made more reliable by pattern matching instead of matching the previous version number?

caoccao:
Interesting... The script I prefer works regardless of what the previous version is so that it can update all of them.

cjee21:
> IIRC it is the case, just that some files were not in the list of files to update and [it was added](https://github.com/MediaArea/MediaArea-Utils/pull/303/files)

For some MSVC2022 projects it is true that it is not in the list and was fixed by that. But for files like MSVC2019 ones and Qt GUI, it was already in the list but it somehow got out of sync/mismatched and required manually fixing in this commit and the ones at MediaInfo. So I think there is room for improvement of the upgrade scripts.

> Script is at https://github.com/MediaArea/MediaArea-Utils/blob/master/MediaInfoLib/UpgradeVersion.sh and we may accept patches if relevant.

MediaInfo one here: https://github.com/MediaArea/MediaArea-Utils/blob/master/MediaInfo/UpgradeVersion.sh If anyone patches the above, this should be patched similarly.

PR #2148, opened 2024-11-25
PNG: adapt cLLi/mdCv letter case

JeromeMartinez:
I am not the only one playing with letter case :).

### DIFF
--- a/Source/MediaInfo/Image/File_Png.h
+++ b/Source/MediaInfo/Image/File_Png.h
@@ -65,11 +65,13 @@ private :
     void IHDR();
     void PLTE() {Skip_XX(Element_Size, "Data");}
     void cICP();
-    void cLLi();
+    void cLLI();
+    void cLLi() { cLLi(); }
     void iCCP();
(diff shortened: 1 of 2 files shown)


wader:
The use of lower/upper-casing to signal stuff is a bit exotic design :)

Source/MediaInfo/Image/File_Png.h
```diff
@@ -65,11 +65,13 @@ private :
     void IHDR();
     void PLTE() {Skip_XX(Element_Size, "Data");}
     void cICP();
-    void cLLi();
+    void cLLI();
+    void cLLi() { cLLi(); }
```
cjee21, line 69: @JeromeMartinez Saw this warning while building:
```
MediaInfoLib\Source\MediaInfo\Image\File_Png.h(69,1): warning C4717: 'MediaInfoLib::F
ile_Png::cLLi': recursive on all control paths, function will cause runtime stack overflow
```
I think you meant 
`void cLLi() { cLLI(); }` with capital `I`.

JeromeMartinez: Oops... Thanks!

PR #2159, opened 2024-12-02
Windows: Enable EH Continuation Metadata

JeromeMartinez:


cjee21:
Previously I did not enable CFG/CET for RegressionTest Release x64.

cjee21:
Oh and there is also MediaInfoShellExt.vcxproj which has CFG/CET enabled for Release but not EHCONT.

JeromeMartinez:
> Previously I did not enable CFG/CET for RegressionTest Release x64.

Not needed, removed.

> Oh and there is also MediaInfoShellExt.vcxproj which has CFG/CET enabled for Release but not EHCONT.

Needed, added.

Thanks for double checking.

cjee21:
@JeromeMartinez and I just discovered for ARM there is `signret` 😮 😆

cjee21:
> @JeromeMartinez and I just discovered for ARM there is `signret` 😮 😆

Looks like this option in Visual Studio is for enabling pointer authentication security mitigation on ARM.

JeromeMartinez:
> Looks like this option in Visual Studio is for enabling pointer authentication security mitigation on ARM.

I'll let you manage that :-p.
(after ARM PRs are merged)

cjee21:
> I'll let you manage that :-p. (after ARM PRs are merged)

Might need to add ARM64EC or ARM64X build if you want to have native ARM MediaInfo.dll for the GUI on Windows 11 ARM64.

JeromeMartinez:
> Might need to add ARM64EC or ARM64X build if you want to have native ARM MediaInfo.dll for the GUI on Windows 11 ARM64.

After the first shot with DLL and CLI.

cjee21:
> CLI.

If you make the CLI fully ARM64, need to build an ARM64 graph plugin for it right?

JeromeMartinez:
> If you make the CLI fully ARM64, need to build an ARM64 graph plugin for it right?

And FFmpeg plugin. But really not the priority, ARM64 is already low usage, ARM64 plugins even lower.

cjee21:
> And FFmpeg plugin. But really not the priority, ARM64 is already low usage, ARM64 plugins even lower.

Or can just make the CLI ARM64EC so that it has native performance but can load x64 plugins.

Microsoft and Qualcomm keeps pushing ARM64 so  if they succeed, usage may increase significantly.

JeromeMartinez:
> Or can just make the CLI ARM64EC so that it has native performance but can load x64 plugins.

Could work in this way? I thought it was ARM64EC = x64 --> ARM64EC, not ARM64EC --> x64.
Anyway, I think I prefer to build all native for the plugins, not the most difficult.
We'll see.
For the moment, really not the priority, CLI is already a bonus, the sponsorship being only on MediaInfo DLL without plugins.

PR #2161, opened 2024-12-04
Add ARM64X build for MSVC2022

cjee21:
Depends on:
- [ ] https://github.com/MediaArea/zlib/pull/21
- [ ] https://github.com/MediaArea/ZenLib/pull/165

Comparison with ARM64:

Screenshot 2024-12-04 123432


cjee21:
> In practice an ARM64X hybrid binary will have about 30% code bloat over a pure classic ARM64 binary

Looks like this does not apply to MediaInfoLib

cjee21:
Seems Windows 11 ARM PCs have about [1.5% market share](https://www.techradar.com/pro/Only-about-720000-Qualcomm-Snapdragon--laptops-sold-since-launch). Probably slightly more than 32-bit Windows which some statistics show as less than 1%.

JeromeMartinez:
> Looks like this does not apply to MediaInfoLib

:(.
So same as macOS "Universal Binaries", just 2x the size in 1 binary, still useful but not so much.
Anyway, it would be easy to switch back to ARM64 only when we get ride of the x64 GUI.

JeromeMartinez:
> about 30% code bloat over a pure classic ARM64 binary [...] Looks like this does not apply to MediaInfoLib

About that, there may be a method for doing a "network" interface instead of having 2 binaries, only launching the library process and doing the API stuff through the local network or similar (there are some memory share stuff), so not depending on the ARM64/EC/X stuff, but for the moment the size is not a big issue so let's already deliver something fully usable (especially the Win11 shell extension and the ToolTip).

PR #2173, opened 2025-01-14
Update .spec file for Fedora 41

g-maxime:


JeromeMartinez:
How difficult would it be to have a `python3-mediainfo` instead of just dropping `python2-mediainfo`?

g-maxime:
> How difficult would it be to have a `python3-mediainfo` instead of just dropping `python2-mediainfo`?

Already here https://github.com/MediaArea/MediaInfoLib/blob/e9153b3f99156a5baad32d0d4e51409fe47795fc/Project/GNU/libmediainfo.spec#L234

JeromeMartinez:
> Already here 

Oops :).

PR #2174, opened 2025-01-17
DAT: fix prioritization based on file extension

JeromeMartinez:


JeromeMartinez:
@dericed I fixed the issue you have with some .hda files, but it is a hot fix, prioritizing the DAT parser over the DV parser when the extension is .hda, so rename the file with .hda extension if you have false positive DV analysis, until I fix the DV parser for not catching some DAT files as DV.

PR #2177, opened 2025-01-24
MP4: Handle `©xyz` location metadata

cjee21:
Metadata tag used by Android phones to store location coordinates for videos recorded by camera.

Before:
```
xyz                                      : +01.5515+110.3584/
```

After:
```
Location                                 : +01.5515, +110.3584
```

MediaTrace:
```
19268D0   Text - Location (30 bytes)
19268D0    Header (8 bytes)
19268D0     Size:                               30 (0x0000001E)
19268D4     Name:                               ©xyz
19268D8    Size:                                18 (0x0012)
19268DA    Language:                            5575 (0x15C7) - eng
19268DC    Value:                               +01.5515+110.3584/
```

Not sure if it is better to format the coordinates for Text/HTML display so that it is easier/nicer to read and can be copy pasted directly to other apps such as Google Maps or to display the original raw string.

### DIFF
--- a/Source/MediaInfo/MediaInfo_Inform.cpp
+++ b/Source/MediaInfo/MediaInfo_Inform.cpp
@@ -56,6 +56,7 @@
 
 //---------------------------------------------------------------------------
 #include <ctime>
+#include <regex>
 //---------------------------------------------------------------------------
 
 namespace MediaInfoLib
@@ -852,6 +853,24 @@ Ztring MediaInfo_Internal::Inform (stream_t StreamKind, size_t StreamPos, bool I
                         if (Valeur.size()>=3 && Valeur.rfind(__T(" / "))== Valeur.size()-3)
                             Valeur.resize(Valeur.size()-3);
                     }
+                }
+
+                //Handling ISO-6709 location data for Text and HTML display
+                bool TextOrHTML = false;
+                #if defined(MEDIAINFO_TEXT_YES) && (defined(MEDIAINFO_HTML_YES) || defined(MEDIAINFO_XML_YES) || defined(MEDIAINFO_CSV_YES))
+                if (Text) TextOrHTML = true;
+                #else
+                TextOrHTML = true;
+                #endif // defined(MEDIAINFO_TEXT_YES) && (defined(MEDIAINFO_HTML_YES) || defined(MEDIAINFO_XML_YES) || defined(MEDIAINFO_CSV_YES))
+                #if defined(MEDIAINFO_HTML_YES)
+                if (HTML) TextOrHTML = true;
+                #endif // defined(MEDIAINFO_HTML_YES)
+                if ((TextOrHTML) && (Get(StreamKind, StreamPos, Champ_Pos, Info_Name) == __T("Recorded_Location"))) {
+                    std::string ISO6709{ Valeur.To_UTF8() };
+                    std::smatch match;
+                    std::regex pattern(R"(([-+]?(?:\d{2}|\d{4}|\d{6})(?:\.\d*)?)([-+](?:\d{3}|\d{5}|\d{7})(?:\.\d*)?)(?:([-+]\d+(?:\.\d*)?)(?:CRS(.+))?)?/)");
+                    if (std::regex_match(ISO6709, match, pattern))
+                        Valeur.From_UTF8(match[1].str() + " " + match[2].str() + " " + match[3].str() + " " + match[4].str());
                 }
 
                 #if defined(MEDIAINFO_XML_YES) || defined(MEDIAINFO_JSON_YES)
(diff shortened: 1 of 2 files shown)


cjee21:
> Geodata is stored according to ISO-6709 standard.

> For historical reasons, any user data start with "\0xA9", must be followed by its assoicated language code.
> - 0x0012: text string length
> - 0x15c7: lang (locale) code: en

Source: https://android.googlesource.com/platform/frameworks/av/+/refs/heads/main/media/libstagefright/MPEG4Writer.cpp#5726

Source/MediaInfo/MediaInfo_Inform.cpp
```diff
@@ -854,6 +855,24 @@ Ztring MediaInfo_Internal::Inform (stream_t StreamKind, size_t StreamPos, bool I
...
+                #endif // defined(MEDIAINFO_TEXT_YES) && (defined(MEDIAINFO_HTML_YES) || defined(MEDIAINFO_XML_YES) || defined(MEDIAINFO_CSV_YES))
+                #if defined(MEDIAINFO_HTML_YES)
+                if (HTML) TextOrHTML = true;
+                #endif // defined(MEDIAINFO_HTML_YES)
+                if ((TextOrHTML) && (Get(StreamKind, StreamPos, Champ_Pos, Info_Name) == __T("Location"))) {
+                    std::string ISO6709{ Valeur.To_UTF8() };
+                    std::smatch match;
+                    std::regex pattern(R"((([-+]?\d{2})(?:\.\d+)?)(([-+]\d{3})(?:\.\d+)?)/)");
+                    if (std::regex_search(ISO6709, match, pattern))
+                        Valeur.From_UTF8(match[1].str() + ", " + match[3].str());
```
JeromeMartinez, line 873: ", " --> " ", it is also supported by Google Maps.
With depth and CRS, I would output e.g. "+27.5916 +086.5640 +8850 WGS_84".

Source/MediaInfo/Multiple/File_Mpeg4.cpp
```diff
@@ -3110,6 +3111,7 @@ File_Mpeg4::method File_Mpeg4::Metadata_Get(std::string &Parameter, int64u Meta)
         case Elements::moov_meta___wrn : Parameter="Warning"; Method=Method_String; break;
         case Elements::moov_meta___wrt : Parameter="Composer"; Method=Method_String; break;
         case Elements::moov_meta___xpd : Parameter="ExecutiveProducer"; Method=Method_String; break;
+        case Elements::moov_meta___xyz : Parameter="Location"; Method=Method_String; break;
```
JeromeMartinez, line 3114: "Location" --> "Recorded_Location" (we already have this MediaInfo field).

Source/MediaInfo/MediaInfo_Inform.cpp
```diff
@@ -854,6 +855,24 @@ Ztring MediaInfo_Internal::Inform (stream_t StreamKind, size_t StreamPos, bool I
...
+                #else
+                TextOrHTML = true;
+                #endif // defined(MEDIAINFO_TEXT_YES) && (defined(MEDIAINFO_HTML_YES) || defined(MEDIAINFO_XML_YES) || defined(MEDIAINFO_CSV_YES))
+                #if defined(MEDIAINFO_HTML_YES)
+                if (HTML) TextOrHTML = true;
+                #endif // defined(MEDIAINFO_HTML_YES)
+                if ((TextOrHTML) && (Get(StreamKind, StreamPos, Champ_Pos, Info_Name) == __T("Location"))) {
+                    std::string ISO6709{ Valeur.To_UTF8() };
+                    std::smatch match;
+                    std::regex pattern(R"((([-+]?\d{2})(?:\.\d+)?)(([-+]\d{3})(?:\.\d+)?)/)");
```
JeromeMartinez, line 871: The regex does not support all ISO6709, and all possibilities are actually found in modern phones, e.g. "+27.5916+086.5640+8850/" and "+27.5916+086.5640+8850CRSWGS_84/".

And your regex does not reject unsupported "3D" coordinates, dangerous, parsing "+27.5916+086.5640+100/" discards the first value and the output is wrong flat coordinates.

Less important (I don't find files with something not "deg"), "deg, min" and "deg, min, sec" styles (digit separator not at the same place) are not supported.

JeromeMartinez (changes requested):
Thank you for the suggestion but correct support is a bit more complicated ;-).
Due to:

Source/MediaInfo/MediaInfo_Inform.cpp
cjee21: I'm not very familiar with regex and did not realize that danger 😓 Learned something new.

I've updated it to handle all (hopefully) ISO6709 now. It gets more complicated 😅 

Looks like [ExifTool's implementation](https://github.com/exiftool/exiftool/blob/e2f9b3a4c57f26eabc82958cb4c6ebe66637c4e2/lib/Image/ExifTool/QuickTime.pm#L8664-L8694) is not fail-proof either.

JeromeMartinez (approved):

PR #2186, opened 2025-02-10
MP4: fix location integer to string

JeromeMartinez:
On some systems the default locale contains "," for decimal separator.
Hot fix, we may use something else than snprintf in the future.

cc @cjee21.


cjee21:
There is `std::format` but using it will make C++20 a requirement.

JeromeMartinez:
> There is std::format but using it will make C++20 a requirement.

Not acceptable :-p.
There is also std::sstream but more complicated, fine with that for the moment.

PR #2193, opened 2025-02-12
Remove printf related warnings

JeromeMartinez:


cjee21:
Some other build warnings remain for Win32: https://github.com/cjee21/MediaInfoLib/actions/runs/13283811466/job/37087892308

JeromeMartinez:
> Some other build warnings remain for Win32

Argh, the 32-bit :). Will fix.

PR #2194, opened 2025-02-12
Migrate MSVC checks from AppVeyor to GitHub Actions

cjee21:


JeromeMartinez:
@cjee21 I just removed the Appveyor hook, please force push with a slight modification e.g. last commit date for forcing the CI to run again.

JeromeMartinez:
With https://github.com/MediaArea/MediaInfoLib/pull/2196 we'll be with all on GitHub and all warnings from all compilers caught.

cjee21:
> all warnings from all compilers caught.

But take note that MSVC MediaInfoLib is still on level 1 warnings. The other projects are at level 3 or 4. Last I checked, even increasing to level 2 produces a number of warnings.

JeromeMartinez:
> Last I checked, even increasing to level 2 produces a number of warnings.

True, but remaining issues with level 2 after my pass today are only possible loss of data due to conversions, I'll manage that... but not this week.
GCC and LLVM use the default warnings.

PR #2201, opened 2025-02-18
Fix alignment of text output for CJK languages

cjee21:
When CJK languages are used, change all ASCII characters and spaces to full width so that they occupy the same space as CJK characters ensuring alignment is maintained.


JeromeMartinez (approved):
Great! (and I learn about Unicode every day...)

Let's try that in the next release.

PR #2208, opened 2025-02-27
Fix UTF-8 micro symbol in string

### LINKED ISSUE #578: File_DolbyE.cpp - a terminal null is present inside a string
V638 A terminal null is present inside a string. The '\0xC2' characters were encountered. Probably meant: '\xC2'. file_dolbye.cpp 847

https://github.com/MediaArea/MediaInfoLib/blob/46c0b99a625aacd3b14fc9389a6e2293788d56b0/Source/MediaInfo/Audio/File_DolbyE.cpp#L847

cjee21:
This also closes #578 

Note that `File_MiXml.cpp` has inconsistent line endings:
mi


JeromeMartinez (approved):

JeromeMartinez:
> has inconsistent line endings

I'll fix that when I do a complete review of such annoyances (it is not the only one).

PR #2212, opened 2025-03-01
AC3: Fix wrong size of mixlevel

StollD:
According to the AC3 specification, the size of mixlevel and mixlevel2 is always five bits, not eight. Parsing it as eight bits will show invalid data, because it will end up including the values of roomtype and the copyright bit. Similarly, the following fields will also not show their real values, but whatever value is written three bits ahead.


JeromeMartinez (approved):
I wonder how it was not caught a lot sooner 🙈.
Thank you a lot of catching that.

PR #2230, opened 2025-03-24
Curl: Fix parsing Literal IPv6 Addresses in URLs

cjee21:
Fixes https://github.com/MediaArea/MediaInfoLib/issues/2233

Hope nothing gets broken by this.

Test:
```
General
Complete name                            : http://[2400:8901::f03c:94ff:fed0:3fee]:80/images-nc/knob_green.png
Format                                   : PNG
Format/Info                              : Portable Network Graphic
File size                                : 9.52 KiB

Image
Format                                   : PNG
Format/Info                              : Portable Network Graphic
Compression                              : Deflate
Format settings                          : Indexed-8
Width                                    : 299 pixels
Height                                   : 299 pixels
Display aspect ratio                     : 1.000
Color space                              : RGB
Compression mode                         : Lossless
Stream size                              : 9.52 KiB (100%)
Gamma                                    : 0.455
```

### DIFF
--- a/Source/MediaInfo/Reader/Reader_libcurl.cpp
+++ b/Source/MediaInfo/Reader/Reader_libcurl.cpp
@@ -88,6 +88,31 @@ namespace Http
             Input           = Input.substr(0, Delimiter_Pos);
         }
     }
+    static void CutWithBracketsColon(std::string& Input, std::string& Output)
+    {
+        size_t BracketCount = 0;
+        const auto Input_Size = Input.size();
+        for (size_t i = 0; i < Input_Size; ++i) {
+            switch (Input[i])
+            {
+            case '[':
+                ++BracketCount;
+                break;
+            case ']':
+                if (!BracketCount)
+                    return;
+                --BracketCount;
+                break;
+            case ':':
+                if (!BracketCount) {
+                    Output = Input.substr(i + 1);
+                    Input.resize(i);
+                    return;
+                }
+                break;
+            }
+        }
+    }
 
     class Url
     {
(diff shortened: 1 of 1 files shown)


Source/MediaInfo/Reader/Reader_libcurl.cpp
```diff
@@ -101,7 +101,12 @@ namespace Http
             CutHead (Host,  User,       "@"     );
             CutTail (Host,  Path,       "/"     );
             CutTail (User,  Password,   ":"     );
-            CutTail (Host,  Port,       ":"     );
+            if (Host.find("]")!=(size_t)-1) {
+                CutTail (Host,  Port,       "]:"    );
+                if (Host.find("]")==(size_t)-1) Host += "]";
+            }
+            else
+                CutTail (Host,  Port,       ":"     );
```
JeromeMartinez, line 109:
```suggestion
            CutWithBracketsColumn(Host, Port);
```

with 
```c++
    static void CutWithBracketsColumn(std::string &Input, std::string &Output)
    {
        size_t BracketCount = 0;
        const auto Input_Size = Input.size();
        for (size_t i = 0; i < Input_Size; i++) {
            switch (Input[i])
            {
            case '[':
                BracketCount++;
                break;
            case ']':
                if (!BracketCount)
                    return;
                BracketCount--;
                break;
            case ':':
                if (!BracketCount) {
                    Output = Input.substr(i + 1);
                    Input.resize(i);
                    return;
                }
                break;
            }
        }
    }
```

After `CutTail`definition.
It seems more error proof e.g. with your code an extra `]` which does not exist at the end is added with some malformed strings, I prefer to avoid "fantom" characters and to put all in `Host` when it is not well understood.

JeromeMartinez:
Thanks.

PR #2241, opened 2025-04-01
Feat: Add display aspect ratio 9:16

### LINKED ISSUE #2240: Display Aspect Ratio not working with 9:16 ratio, but works with FFprobe
There seems to be an issue with the "Display Aspect Ratio" field in MediaInfo when working with videos that have a 9:16 aspect ratio. While FFprobe correctly displays the 9:16 aspect ratio, MediaInfo fails to do so.

Steps to reproduce:

Analyze a video with a 9:16 aspect ratio using MediaInfo.

Check the "Display Aspect Ratio" field, which will not show the correct value (9:16) but 0.562

Analyze the same video using FFprobe, and it will correctly show the 9:16 ratio.

If I swap the width and height values manually (i.e., replace the width with the height and vice versa), MediaInfo will then correctly display the ratio as 16:9, which suggests there may be an issue with how MediaInfo is handling aspect ratio calculations for 9:16.

Expected result:

The "Display Aspect Ratio" field should display the correct 9:16 aspect ratio for videos with this ratio, as FFprobe does.

Actual result:

The "Display Aspect Ratio" field in MediaInfo fails to correctly display the 9:16 aspect ratio.

Additional information:

The issue appears specifically with the 9:16 aspect ratio.

Swapping width and height results in the correct 16:9 ratio in MediaInfo.

xavierlaffargue:
Fix #2240 
Add display aspect ratio 9:16


JeromeMartinez:
Thanks.

Issue #2242, opened 2025-04-03
Please bump minimum Cmake requirement from 3.1 (to at least 3.5)


anjos:
I'm maintaining the [conda-forge builds for libmediainfo](https://github.com/conda-forge/libmediainfo-feedstock).

In one of our latest builds it came to my attention that the cmake files from this project are maybe a bit outdated and would require a small change for better compatibility.  Our build stack uses `cmake==4.0.0` and it complains like so, [halting the build](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1214802&view=logs&jobId=7b6f2c87-f3a7-5133-8d84-7c03a75d9dfc&j=7b6f2c87-f3a7-5133-8d84-7c03a75d9dfc&t=9eb77fd2-8ddd-5444-8fc0-71cb28dcb736):

```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.


-- Configuring incomplete, errors occurred!
```

For now, I patched the build as recommended with `-D CMAKE_POLICY_VERSION_MINIMUM=3.5`, and this seems to work well across different platforms. However it would seem better if you modified the top-level CMakeList.txt to either set this value to (at least) 3.5 or further up if that makes sense.

JeromeMartinez:
> I'm maintaining the [conda-forge builds for libmediainfo](https://github.com/conda-forge/libmediainfo-feedstock).

Thanks for doing that.

> cmake files from this project are maybe a bit outdated

Not really outdated, we update it when we add a new parser (so a new C++ file), just that we don't use fancy stuff in the CMake file so we keep the compatibility with old platforms as much as possible.

> would require a small change for better compatibilit

We stopped to support very old distro versions like CentOS 7 last year, CentOS 8 has CMake 3.11 and Ubuntu 20.04 has CMake 3.16, and it looks like that our CMake file has no problem with CMake 3.5 in your tests so no high cost of upgrading, in addition to the new CMake 4.0 doing an error by default, it seems legitimate to do that, we'll do it if nobody complains.

JeromeMartinez:
[...]

PR #2243, opened 2025-04-03
Bump CMake min version to 3.5

JeromeMartinez:
In order to have CMake 4.0 working without option

Fix https://github.com/MediaArea/MediaInfoLib/issues/2242.

### DIFF
--- a/Project/CMake/CMakeLists.txt
+++ b/Project/CMake/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1.0)
+cmake_minimum_required(VERSION 3.5.0)
 
 project(MediaInfoLib)
 


JeromeMartinez:
@g-maxime I don't like the warning:

```
CMake Warning (dev) at /home/runner/work/MediaInfoLib/ZenLib/Project/CMake/CMakeLists.txt:13 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  For compatibility with older versions of CMake, option is clearing the
  normal variable 'BUILD_SHARED_LIBS'.
```

Maybe there is a way to avoid a risk in a change of behavior when we bump the minimal version to 3.13+?

Issue #1391, opened 2021-04-02
Support WebP format


erri120:
Links:
- [Wikipedia](https://en.wikipedia.org/wiki/WebP)
- [Overview](https://developers.google.com/speed/webp/)
- [Compression Techniques](https://developers.google.com/speed/webp/docs/compression)
- [libwebp API Docs](https://developers.google.com/speed/webp/docs/api)
- [libwebp Source Code](https://chromium.googlesource.com/webm/libwebp/)

PR #2262, opened 2025-04-16
WebP support

bitplane:
Hi, I noticed that WebP didn't work, so thought I'd add it using the info from #1391 

General file info is still handled by the RIFF parser, this just adds basic image stream support. It doesn't rely on libwebp, to keep the binary sizes down. It doesn't do EXIF metadata (that'd need the JPEG EXIF code split out?). I tried to add the .cpp files to all the projects, but only built it on Linux.

Test data with vp8, lossless, alpha channel, animated and iccp colour profiles attached:

[testdata.zip](https://github.com/user-attachments/files/19781107/testdata.zip)

### DIFF
--- a/Source/MediaInfo/Image/File_WebP.cpp
+++ b/Source/MediaInfo/Image/File_WebP.cpp
@@ -1,0 +1,359 @@
+/*  Copyright (c) MediaArea.net SARL. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license that can
+ *  be found in the License.html file in the root of the source tree.
+ */
+
+//---------------------------------------------------------------------------
+// Pre-compilation
+#include "MediaInfo/PreComp.h"
+#ifdef __BORLANDC__
+    #pragma hdrstop
+#endif
+//---------------------------------------------------------------------------
+
+//---------------------------------------------------------------------------
+#include "MediaInfo/Setup.h"
+//---------------------------------------------------------------------------
+
+//---------------------------------------------------------------------------
+#if defined(MEDIAINFO_WEBP_YES)
+//---------------------------------------------------------------------------
+
+//---------------------------------------------------------------------------
+#include "MediaInfo/Image/File_WebP.h"
+#if defined(MEDIAINFO_ICC_YES)
+    #include "MediaInfo/Tag/File_Icc.h"
+#endif
+#if defined(MEDIAINFO_XMP_YES)
+    #include "MediaInfo/Tag/File_Xmp.h"
+#endif
+#if defined(MEDIAINFO_VP8_YES)
+    #include "MediaInfo/Video/File_Vp8.h"
+#endif
+using namespace std;
+//---------------------------------------------------------------------------
+
+namespace MediaInfoLib
+{
+
+//***************************************************************************
+// Const
+//***************************************************************************
+
+namespace Elements
+{
+    const int32u RIFF = 0x52494646;
+    const int32u WEBP = 0x57454250;
+    const int32u WEBP_ALPH = 0x414C5048;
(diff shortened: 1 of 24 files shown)


JeromeMartinez:
Thank you for the proposal.

I adapted it for being more in MediaInfo style.
Also added right display of [lossless with alpha](https://developers.google.com/speed/webp/gallery2).
RIFF parser is similar to the one in File_Riff.cpp, maybe in the future it would deserve that I merge codes.

Also found that WebP is a specific image format only for lossless content
First update, I'll do another one tomorrow, but @bitplane please check and comment.

>  iccp colour

No ICC in vp8x_iccp file.

> (that'd need the JPEG EXIF code split out?).

Yes!

JeromeMartinez:
I added PRs for adapting the VP8 parser so it can accept image tracks, then rebased your branch.
I tweaked the code for matching MediaInfo fields e.g. "alpha" goes to colorspace ("YUVA" instead of "YUV"), I disabled fields you added but not usually in MediaInfo, it could be added if there is a demonstrated need for it.
I switched "Animation" WebP to a video stream in order to use frame count, duration, and so on (should be done for GIF too, but not yet implemented).
I linked the ICC and XMP parsers (note: XMP parsing is not full, I tweaked it for catching the GIMP creation tool in the file I have, but it would deserve more work, on my side I plan to work on that only after I switch from tinyXML2 parser to the internal one) to the corresponding WebP chunks, nothing for EXIF because I currently have no EXIF parser (if you are motivated... ;-) just that we don't accept external libs so it should be native implementation)
I use VP8/VP8L width/height, never the "canvas" fields found elsewhere, looks like that they are less important.

bitplane:
Awesome, thank you! Sorry about making this work for you! I just couldn't remember whether I'd resized a picture before converting to webp, `mediainfo` didn't work with it, so I waded in without much thought or research and added it!

bitplane:
>  (if you are motivated... ;-)

Maybe! I know very little about EXIF, but I'll have a look if I can find time. Should I open an issue for it, and add discussion / notes there?

cjee21:
Having EXIF parsing and c2pa detection in MediaInfo would be nice :)

JeromeMartinez:
> Should I open an issue for it, and add discussion / notes there?

A discussion could be in a dedicated PR (you can open it even if it is a draft), else a new issue ticket.

cjee21:
> @cjee21 are you motivated? ;-).

I think I can contribute and help test if someone lays the groundwork. Not sure how to start from scratch.

JeromeMartinez:
> I think I can contribute and help test if someone lays the groundwork. Not sure how to start from scratch.

It is very similar to the IAMF PR you just did. I can start them if you feel less confortable than with IAMF, but not before next week on my side.

cjee21:
> It is very similar to the IAMF PR you just did.
I don't even know if I did it correctly 😄 

> not before next week
I have enough PRs open right now, don't feel like starting another one for now 😅 
Meanwhile maybe @bitplane start on it

Source/MediaInfo/Image/File_WebP.cpp
```diff
@@ -0,0 +1,359 @@
...
+{
+    //Parsing
+    #if defined(MEDIAINFO_XMP_YES)
+    File_Xmp MI;
+    Open_Buffer_Init(&MI);
+    Open_Buffer_Continue(&MI);
+    Open_Buffer_Finalize(&MI);
+    Merge(MI, Stream_General, 0, 0);
+    #else
+    Skip_UTF8(Element_Size                                      "XMP metadata");
```
cjee21, line 353: @JeromeMartinez Missing comma

PR #2269, opened 2025-04-23
XMP: support xmp:CreatorTool & misc

JeromeMartinez:


JeromeMartinez:
> @JeromeMartinez I managed to get the XMP part for JPEG but how to send it to XMP parser? When I tried to parse the XMP XML with File_Xmp, the XMP section disappears from trace.

Have a look at the [WebP related implementation](https://github.com/MediaArea/MediaInfoLib/blob/master/Source/MediaInfo/Image/File_WebP.cpp#L343-L355).

JeromeMartinez:
> Update: @JeromeMartinez I see what is happening now, it is working but just disappears from trace output.

Parsing of XML is not really compatible with the trace feature. On my todo-list as a more global way to show XML in MediaTrace output (when we switch to our internal XML parser), not in the scope of the support of XMP or other XML related format in any container.

JeromeMartinez:
> Some things disappear, is it okay?

No, there is a mix up between tracks.
Provide a PR and I review/fix.

PR #2280, opened 2025-04-27
ICC: Display description (Profile name for display)

cjee21:
Useful to determine if JPEGs are sRGB, Adobe RGB or Display P3.


cjee21:
How is this PR?

JeromeMartinez:
On hold for when I have more time for commenting.
In short: sRGB, Adobe RGB or Display P3 are color primaries and any tip about it e.g. in ICC description should be mapped to color_primaries field and not to a new field.

JeromeMartinez:
> That is why in the first place I just show profile description as a new/separate field so that whatever it is, at least user knows what kind of profile is included.

I started this kind of display with a quick "ColorSpace_ICC" but I wonder if it is good to go in this direction... Maybe.
Would it make sense to currently comment your patch (keeping tips but not using it now) and use "ColorSpace_ICC" using the description as a first step?

cjee21:
Show profile description as 'ColorSpace_ICC' ?

cjee21:
By the way, IIRC, I have seen images with uRGB profile. A minified version of sRGB maybe?

JeromeMartinez:
> By the way, IIRC, I have seen images with uRGB profile. A minified version of sRGB maybe?

micro, or nano, there are so many flavors of ICC profiles. Reasons I don't really know what to do there, a least without catching how to categorize profiles, and description fields may also be a bit misleading. a long topic here and I don't have the time and the motivation for that for the moment.

"colour_primaries_ICC_Description" for the MediaInfo name, maybe, as we don't check "colour_primaries values in ICC, actually.

PR #2290, opened 2025-05-11
MP4: Map more tags

### LINKED ISSUE #1590: Support Publisher metadata in MP4 tags
In [`File_Mpeg4.cpp`](https://github.com/MediaArea/MediaInfoLib/blob/master/Source/MediaInfo/Multiple/File_Mpeg4.cpp), there is no support for the `©pub` and `©nrt` boxes holding _publisher_ and _narrator_ information, respectively. [Other less common fields](https://community.mp3tag.de/t/mp4-felder-atom-box-in-anderer-software/48614/7?u=crissov) are probably unsupported, too. 

PS: It would be nice if MP4 tag support was properly [documented on the homepage](https://mediaarea.net/en/MediaInfo/Support/Tags) as well. Also, that page (like the source code repo) is too hard to find there.

cjee21:
Map more Arranger, Label, Lyricist and Publisher tags in MP4.
Fix uninitialized variables warnings for File_Mpeg4.cpp.

Resolves #1590

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mpeg4.cpp
+++ b/Source/MediaInfo/Multiple/File_Mpeg4.cpp
@@ -89,6 +89,7 @@ namespace Elements
     const int64u moov_meta___prd=0xA9707264;
     const int64u moov_meta___PRD=0xA9505244;
     const int64u moov_meta___prf=0xA9707266;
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
Thanks.

PR #2293, opened 2025-05-12
HTML: Display cover image

cjee21:
At the moment only usable with CLI:

```
MediaInfo --cover_data=base64 --output=HTML <filename> > out.html
```

validated for when single file is opened:

Screenshot 2025-05-12 at 15-40-58 Showing results for uploaded file test html - Nu Html Checker

### DIFF
--- a/Source/MediaInfo/MediaInfo_Inform.cpp
+++ b/Source/MediaInfo/MediaInfo_Inform.cpp
@@ -454,7 +454,7 @@ Ztring MediaInfo_Internal::Inform()
 
     if (HTML)
         Retour+=__T("<!DOCTYPE html>\n"
-            "<html>\n"
+            "<html lang=\"") + MediaInfoLib::Config.Language_Get(__T("  Language_ISO639")) + __T("\">\n"
             "  <head>\n"
             "    <meta charset=\"utf-8\">\n"
             "    <meta name=\"viewport\" content=\"width=device-width, initial-scale=1\">\n"
@@ -467,6 +467,29 @@ Ztring MediaInfo_Internal::Inform()
             "      table { width: 100%; padding: 2px; border-spacing: 2px; border:1px solid var(--border-color, navy); margin-bottom: 18px; }\n"
             "      td.Prefix { width: 25vw; min-width: 150px; max-width: 500px; font-style: italic; }\n"
             "      h2 { margin-top: 0px; }\n"
+            "      .cover-image {\n"
+            "        height: 100px;\n"
+            "        width: auto;\n"
+            "        cursor: pointer;\n"
+            "      }\n"
+            "      .overlay-sheet {\n"
+            "        display: none;\n"
+            "        position: fixed;\n"
+            "        top: 0;\n"
+            "        left: 0;\n"
+            "        width: 100%;\n"
+            "        height: 100%;\n"
+            "        background-color: rgba(0, 0, 0, 0.8);\n"
+            "        justify-content: center;\n"
+            "        align-items: center;\n"
+            "        z-index: 1000;\n"
+            "        cursor: pointer;\n"
+            "      }\n"
+            "      .overlay-image {\n"
+            "        max-width: 90vw;\n"
+            "        max-height: 90vh;\n"
+            "        cursor: pointer;\n"
+            "      }\n"
             "    </style>\n"
             "  </head>\n"
             "  <body>\n");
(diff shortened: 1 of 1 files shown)


Source/MediaInfo/MediaInfo_Inform.cpp
```diff
@@ -934,6 +976,19 @@ Ztring MediaInfo_Internal::Inform (stream_t StreamKind, size_t StreamPos, bool I
...
                     Retour+=Valeur;
                     Retour+=__T("</td>\n      </tr>");
+
+                    #if MEDIAINFO_ADVANCED
+                    if (MediaInfoLib::Config.Flags1_Get(Flags_Cover_Data_base64)) {
+                        if (Get((stream_t)StreamKind, StreamPos, Champ_Pos, Info_Name) == __T("Cover")) {
+                            Retour += __T("\n      <tr>\n        <td class=\"Prefix\">");
+                            Retour += __T("Cover image :</td>\n        <td>");
+                            Retour += __T(R"(<img class="cover-image" alt="Cover image preview" src=")");
+                            Retour += __T("data:") + Get((stream_t)StreamKind, StreamPos, __T("Cover_Mime")) + __T(";base64,") + Get((stream_t)StreamKind, StreamPos, __T("Cover_Data"));
```
JeromeMartinez, line 986: `Cover_Mime` and `Cover_Data` may have multiple contents separated by " / ", this code works when there is a unique cover but would fail when there are multiple covers, you need to keep only what is before " / ".

JeromeMartinez (changes requested):
Great!
One small change request:

Source/MediaInfo/MediaInfo_Inform.cpp
cjee21: Then we need to split the covers so that can display all of them?

JeromeMartinez: > Then we need to split the covers so that can display all of them?

I wonder if it is good or not to show all covers (front, back, cd, usually) in the HTML view.
My first shot would be to show only the first cover and ignoring the others in order to not have too many things in the HTML view, but it is not a strong opinion. (option for 1 or all covers? It becomes to many options maybe...)

cjee21: Then I'll just make it display first only for now. I think all my files only have one cover too.

cjee21:
https://github.com/user-attachments/assets/d7af59c7-eb5a-4b56-9aa5-844d1e9aad74

Source/MediaInfo/MediaInfo_Inform.cpp
JeromeMartinez: Checking the video you sent and how it is displayed, I am now in favor of having all covers displayed, what do you think?

cjee21: I think just show all. (I just discovered Windows shows back cover as thumbnail for the file below.)

Screenshot 2025-05-13 at 14-20-22 MediaInfo

But we now have another issue. After I rebased on master, cover display no longer works and in some cases even an empty (0 bytes) HTML output with `--cover_data=base64` enabled.

JeromeMartinez: > After I rebased on master, cover display no longer works

the "API" should not have changed so a regression from my PR yesterday, I check.

cjee21: Something gone wrong it seems...

Before:
Screenshot 2025-05-13 at 14-36-17 file ___C _Users_cjees_Desktop_HTML%20cover%20art_test html

After:
Screenshot 2025-05-13 at 14-36-58 file ___D _Software_Development_MediaInfo_MediaInfo_Project_MSVC2022_x64_Debug_test html

This is for the same file, a FLAC file.

cjee21: At the moment I force push with multiple cover display and without rebase.

JeromeMartinez: Tested with 3 and 4 covers, fine.

JeromeMartinez (approved):

cjee21:
I'll wait for 2299 to be merged then rebase this and c2pa etc.

cjee21:
Okay, tested with a few files of various formats and it displays fine. I can start looking at the GUI side regarding this.

PR #2295, opened 2025-05-12
When outputting JSON, return an array of objects…

trallen:
…even if only one file is requested.

This allows for consistent handling of JSON output in cases of arbitrary numbers of input files.

Fixes MediaArea/MediaInfo#1103


trallen:
Thanks @JeromeMartinez. I understand; please let me know what changes you determine are needed!

alex3kov:
> …even if only one file is requested.
> 
> This allows for consistent handling of JSON output in cases of arbitrary numbers of input files.
> 
> Fixes [MediaArea/MediaInfo#1103](https://github.com/MediaArea/MediaInfo/issues/1103)

Tested - works as suggested in issue.

PR #2299, opened 2025-05-13
Add image track for thumbnails and covers, fix decode

JeromeMartinez:
Fix after https://github.com/MediaArea/MediaInfoLib/pull/2289


JeromeMartinez:
@cjee21 it should fix the issue with cover in base64.

cjee21:
> @cjee21 it should fix the issue with cover in base64.

Looks good after testing a few FLAC and MKV along with the cover art display PR.

PR #2300, opened 2025-05-13
JNA: Fix Java to Native integer conversion

g-maxime:
Fixes MediaArea/MediaInfo#521


JeromeMartinez (approved):
A lot nicer than what I did 20 years ago :).
(I am pretty sure that it was not there at this moment, anyway now we have that)

PR #2301, opened 2025-05-17
build: set cxx std to 17 in cmake 

trim21:


Project/CMake/CMakeLists.txt
```diff
@@ -2,6 +2,9 @@ cmake_minimum_required(VERSION 3.5.0)
 
 project(MediaInfoLib)
 
+set(CMAKE_CXX_STANDARD 17)
```
JeromeMartinez, line 5: We try to keep the maximum compatibility when we have no big reason to remove old systems, for the moment we are still with C++11.

PR #2302, opened 2025-05-17
CMake: Clean-up + C++17 + enable Graphviz runtime

cjee21:


Project/CMake/CMakeLists.txt
```diff
@@ -434,7 +432,7 @@ else()
 endif()
 
 set_target_properties(mediainfo PROPERTIES
-  CXX_STANDARD 11
+  CXX_STANDARD 17
```
JeromeMartinez, line 435: Except if there is a strong need about that (which one?), C++11 should stay there.

JeromeMartinez (changes requested):

Project/CMake/CMakeLists.txt
cjee21: Was just trying to remove the warnings about fallthrough annotations requiring C++17 but looks like it fails to build on Ubuntu with C++17.

JeromeMartinez: > Was just trying to remove the warnings about fallthrough annotations requiring C++17

True that annotation provides a warning with old systems.
Either a #define (need to replace the annotation by a define everywhere and set the define to the annotation or not depending of the C++ version) depending of the C++ version enabled, or maybe there is a compiler option?

Also, the best here IMO is to put CXX_STANDARD depending on the compiler support (if the compiler supports C++17, it would be great that the compilation is in C++17 for optimizations), if it is possible.

cjee21: File_Dat does not build in C++17 on Ubuntu: https://github.com/MediaArea/MediaInfoLib/actions/runs/15086685415/job/42410091788

cjee21: I think just leave it like this for now. Those warnings only appear on Windows it seems. On Windows recommended to use the MSVC project which has flags configured.

JeromeMartinez (approved):

PR #2309, opened 2025-05-26
Location: format all to ° + direction

JeromeMartinez:
ISO 6709 is nice because it is a standard but in practice it is not accepted by e.g. Google Maps, so we move all to something more accepted in practice.
No conversion of floating point grads to grads/minutes/seconds as floating point grads seem to be accepted in different places, so we keep the precision from the source (especially when it from text based tags).

cc @cjee21.

### DIFF
--- a/Source/MediaInfo/File__Analyze_Streams.cpp
+++ b/Source/MediaInfo/File__Analyze_Streams.cpp
@@ -41,119 +41,204 @@ using namespace std;
 namespace MediaInfoLib
 {
 
-//---------------------------------------------------------------------------
-extern MediaInfo_Config Config;
-const char* Mpegv_colour_primaries(int8u colour_primaries);
-//---------------------------------------------------------------------------
+    //---------------------------------------------------------------------------
+    extern MediaInfo_Config Config;
+    const char* Mpegv_colour_primaries(int8u colour_primaries);
+    //---------------------------------------------------------------------------
 
-//***************************************************************************
-// Others, specialized, parsing
-//***************************************************************************
+    //***************************************************************************
+    // Others, specialized, parsing
+    //***************************************************************************
 
-//---------------------------------------------------------------------------
-bool Location_Adapt(string& Value)
-{
-    // Zero padding of degrees/minutes/seconds
-    auto AddZeroDMS = [](string& Value, size_t& degree_pos) {
-        degree_pos = Value.find("\xC2\xB0", degree_pos);
-        if (degree_pos == string::npos) {
-            return false;
-        }
-        auto space_pos = Value.find(' ', degree_pos);
-        bool Modified = false;
+    //---------------------------------------------------------------------------
+    bool Location_Adapt(string& Value)
+    {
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
> and since already formatted here, the quick formatting I did for text and HTML in Inform.cpp should be removed.

True... and it was not working, actually, in the tests I did (I merged by reading that the code does not hurt, but didn't test it), weird, anyway I do like that for the moment...

>  ISO 6709 (there are some requirements for leading zeroes if wiki is correct)

The example is "EXAMPLE 1 50°40'46.461"N 95°48'26.533"W 1,123.45m" so no leading 0 for the degrees (spec says "where the minute or second value is less than 10, the value should include the leading 0"), and I show only degrees, so compatible.

> Well, ISO 6709 has [...] XML [...] representations

I see, with a tuple, I wonder if we should do that (if it widely used), I have no strong opinion but I am switching to the "human interface" representation everywhere for the moment. I may change my mind later (but... What about e.g. JSON in that case?)

cjee21:
> so no leading 0 for the degrees (spec says "where the minute or second value is less than 10, the value should include the leading 0")

I was referring to the Exif output being close to the human representation suggestion. Current Exif output does not pad with zeros for minute and second.

cjee21:
Screenshot 2025-05-27 015142

All my JPEGs have minute and second in Exif data?

cjee21:
Line 130 has 'redundant code'. The condition is repeated.

cjee21:
Actually, I just found the bug. It is on line 109 of File_Analyze_streams and affect all ISO 6709 latitudes that has leading zeroes, not only MP4. No need to change loci tag handling as that encodes in ISO 6709 string which requires leading zeroes.

Source/MediaInfo/File__Analyze_Streams.cpp
```diff
@@ -50,6 +50,102 @@ const char* Mpegv_colour_primaries(int8u colour_primaries);
...
+
+    if (lat_sign_pos || lon_sign_pos == std::string::npos || slash_pos == std::string::npos)
+        return;
+
+    auto lat_sign = Value[lat_sign_pos] == __T('+') ? __T('N') : __T('S');
+    ++lat_sign_pos;
+    while (Value[lat_sign_pos] == __T('0') && Value[lat_sign_pos + 1] >= __T('0') && Value[lat_sign_pos + 1] <= __T('9')) {
+        ++lat_sign_pos;
+    }
+    auto lat_without_sign = Value.substr(1, lon_sign_pos - 1);
```
cjee21, line 109:
```C++
    auto lat_without_sign = Value.substr(lat_sign_pos, lon_sign_pos - lat_sign_pos);
```

JeromeMartinez:
Changed where I adapt the string in order to be able to catch duplicate locations.
There is still a duplicate when precision is not same, I check if I can quickly catch that (what is the real precision in the case of `loci` vs a string? No idea...)

PR #2311, opened 2025-05-27
Location: AddZeroDMS fix

cjee21:
Fix after #2309 

Discovered that a file with single digit minute coordinate was not zero-padded.


JeromeMartinez:
> Maybe it was after the change from Ztring to string in previous PR 

I guess so (byte size of the degree character).

>  I did not test this file after that.

I'll see if I find a file with that when I add more test files and if not I'll craft one, in order to avoid such mistake...

PR #2312, opened 2025-05-28
Exif: Format datetime in ISO 8601 format

cjee21:
Format datetime from Exif in ISO 8601 format so that `DateTime_Adapt` in `File__Analyze_Streams.cpp` can handle it.


JeromeMartinez:
Sorry, I was doing the same about date and it was in QA, I just created https://github.com/MediaArea/MediaInfoLib/pull/2313.
I put it generic because Exif is not the only one have that e.g. DPX too.

Please use this PR for the other commit only.

Source/MediaInfo/File__Analyze_Streams_Finish.cpp
```diff
@@ -1137,6 +1137,10 @@ void File__Analyze::Streams_Finish_StreamOnly_General(size_t StreamPos)
         if (!CompanyName.empty() && Model.rfind(CompanyName + __T(' '), 0) == 0) {
             Fill(Stream_General, StreamPos, General_Encoded_Hardware_Model, Model.substr(CompanyName.size() + 1), true);
         }
+        if (CompanyName == __T("NIKON CORPORATION") && Model.rfind(__T("NIKON "), 0) == 0) {
+            Fill(Stream_General, StreamPos, General_Encoded_Hardware_CompanyName, Ztring("Nikon"), true);
```
JeromeMartinez, line 1141:
```suggestion
            Fill(Stream_General, StreamPos, General_Encoded_Hardware_CompanyName, "Nikon", true, true);
```

Better to use char* version when possible (and it will be easier for us when we switch to UTF-8 in our codebase).

cjee21: Need to include the size else it won't build due to multiple matching overloads.

Source/MediaInfo/File__Analyze_Streams_Finish.cpp
```diff
@@ -1137,6 +1137,10 @@ void File__Analyze::Streams_Finish_StreamOnly_General(size_t StreamPos)
         if (!CompanyName.empty() && Model.rfind(CompanyName + __T(' '), 0) == 0) {
             Fill(Stream_General, StreamPos, General_Encoded_Hardware_Model, Model.substr(CompanyName.size() + 1), true);
         }
+        if (CompanyName == __T("NIKON CORPORATION") && Model.rfind(__T("NIKON "), 0) == 0) {
+            Fill(Stream_General, StreamPos, General_Encoded_Hardware_CompanyName, "Nikon", static_cast<size_t>(-1), true, true);
```
JeromeMartinez, line 1141:
```suggestion
            Fill(Stream_General, StreamPos, General_Encoded_Hardware_CompanyName, "Nikon", Unlimited, true, true);
```

You were right but use the shortcut 'Unlimited' (I should do the static_cast on it, I know...) so it is as elsewhere.

PR #2314, opened 2025-05-28
Fix Fill() called incorrectly with utf8=false parameter

g-maxime:
This is the possible cause of  a crash on Android due to the call of Ztring::From_Local() instead of Ztring::From_UTF8()


JeromeMartinez:
I don't see why it would provoke a crash, but it is good to fix that anyway.
Rebase your branch and apply this fix to the same kind of code (new code) below the faulty line.

PR #2315, opened 2025-05-29
Exif: Display more tags

cjee21:
I have added the common tags such as exposure settings.


cjee21:
Likely needs more tests with images from various cameras.

Known issues:
Tags from Apple devices have many decimal places:
```
Recorded location                        : 37°30'52.68"N 127°07'37.86"E 52.57939530m
Writing application                      : Adobe Photoshop 21.1 (Macintosh)
Writing hardware                         : Apple iPhone XS
LensModel                                : iPhone XS back dual camera 4.25mm f/1.8
LensZoomActualFocalLength                : 4.25000000 mm
LensZoom35mmStillCameraEquivalent        : 26 mm
```

JeromeMartinez (changes requested):
Thanks. I add change requests and:

> Tags from Apple devices have many decimal places:

We show what the file says, if Apple wants such denominator... At long term I think I'll support N/D format as an option, as I see that Apple has denominators like "699009" (why???)

Source/MediaInfo/Tag/File_Exif.cpp
```diff
@@ -1174,6 +1256,17 @@ void File_Exif::Get_IFDOffset(int8u KindOfIFD)
 //---------------------------------------------------------------------------
 void File_Exif::GetValueOffsetu(ifditem &IfdItem)
 {
+    auto GetDecimalPlaces = [](int32u denominator, int8u defaultDecimalPlaces = 3) -> int8u {
```
JeromeMartinez, line 1259: I change my mind there, as there are so many different denominators, and you wish to have up to 9 decimals, it does not work well with e.g. 111. Let's do that if no problem seen on your side:

```suggestion
    auto GetDecimalPlaces = [](int32u denominator) -> int8u {
        int8u count = 0;

        while (denominator > 1) {
            ++count;
            denominator /= 10;
        }

        return count;
    };
```

JeromeMartinez (changes requested):

Source/MediaInfo/Tag/File_Exif.cpp
```diff
@@ -1174,6 +1297,19 @@ void File_Exif::Get_IFDOffset(int8u KindOfIFD)
...
+        if (denominator == 1)
+            return 0;
+        int8u count{ 1 };
+        while (denominator > 10) {
+            ++count;
+            denominator /= 10;
+        }
+        if (denominator == 10)
+            return count;
+        return count + 3;
```
JeromeMartinez, line 1310:
```suggestion
        return count;
```

I don't see why you want more in case of no multiple of 10, with my change I have e.g.:
- 282581/2581 = 109.4851
- 282582/2581 = 109.4855

So we don't lose any precision, no need of extra digit.

cjee21: The example you give is `count + 1`.

JeromeMartinez: > The example you give is count + 1.

the +1 is done in the loop already.

cjee21: > the +1 is done in the loop already.

Right....

> Need to manage cases of more than 1 and less than 1. Altitude with more precision is not useful.

Need a way to handle all the different shutter speeds so that the original fraction can be recovered. Not all cameras use 1/x too...

JeromeMartinez: > Looks okay now? Except focal length 17/4 = 4.2 instead of 4.25

I am in favor to start with such kind of method then we tweak step by step.

JeromeMartinez:
I refactored for having an easier way to extend (I hope), please check that I introduced no issue (I still don't have all files)
e.g. I didn't test SensitivityType part.
Fields order is changed, by Exif tag, but should not be an issue (and order may be forced if we flag a field name in the list of known field names, which will be done for some of the fields e.g. LensMake & LensModel should be in "Lens" know name and like Make & Model i.e. Writing hardware and sub elements).
ExifVersion is hedden in default display (not an information for the file itself), trying to reduce the count of fields in default view.

cjee21:
Added all the version numbers.

cjee21:
> check that I introduced no issue

Well the special handling for Xiaomi having FocalLengthIn35mmFilm = 0 is gone. Up to you if you still want that.

cjee21:
> @cjee21 I nearly clicked on "merge" when you pushed an update :-p, should I wait a bit?

Should be okay now. I just corrected some types to match their actual Exif size/type.

Source/MediaInfo/Tag/File_Exif.cpp
```diff
@@ -815,56 +984,347 @@ void File_Exif::Streams_Finish()
...
+        currentIFD = Kind_GPS;
+        const auto& Infos_GPS = Infos_GPS_It->second;
+        for (const auto& Item : Infos_GPS) {
+            size_t Parameter = 0;
+            Ztring Value;
+            const char* ParameterC = nullptr;
+            string ParameterS;
+            switch (Item.first) {
+            case IFDGPS::GPSVersionID: {
+                ParameterC = "GPSVersionID";
```
JeromeMartinez, line 1252:
```suggestion
                ParameterC = "ExifGPSVersion";
```

ID is not really relevant, and we need to indicate that this is from Exif.

JeromeMartinez (changes requested):
A last tiny change request.

JeromeMartinez (approved):

cjee21:
One final type correction.

PR #2317, opened 2025-05-31
JPEG: Parse MPF

cjee21:


cjee21:
Work-in progress. Just enough to view the IFD.

Todo: Update non-MSVC project files, parse MP Entry etc.

JeromeMartinez:
Why not adding stiff to the Exif parser?I prefer to avoid many files.

JeromeMartinez:
Sorry, my mistake, I was thinking to a MPF Ifd In Exif parser.

cjee21:
Merged with Exif.

JeromeMartinez:
WIP pushed with some modifications with how I see the handling of MPF.
(not the best code but I try to also avoid too many #include)

cjee21:
I split into a separate commit for improving the Exif parsing codes before integrating MPF.

I also fixed MPF parsing and the compiler errors/warnings.

cjee21:
Fixed parsing multiple images.

cjee21:
Next is to handle the multiple images. Right now it only shows one image stream and with a size larger than the full filesize. I do not know how to do that.

cjee21:
It seems what is filled by MPF parsing (if uncomment that line to fill all tags) does not appear in text output.

JeromeMartinez:
Afk for a couple of days but I manage that when I am back.

JeromeMartinez:
@cjee21 I tweaked again your patch, I added the display of MPF images in the MediaInfo report, with type and muxing mode.
Panorama may deserve a merge of the image tracks as it is a single image in practice but I have no file and more work, maybe later.
please review/force-push.

cjee21:
I filled the type of the primary image. Looks okay other than fullparsing mode does not display properly.

JeromeMartinez:
> Looks okay other than fullparsing mode does not display properly.

should be OK now.

cjee21:
> should be OK now

Yes it is.

PR #2318, opened 2025-05-31
Exif: Fixes, improvements and display more data

cjee21:
Code improvements, bug fixes, some Exif 3.0 support and display ImageTitle tag.

### DIFF
--- a/Source/MediaInfo/File__Analyze_MinimizeSize.h
+++ b/Source/MediaInfo/File__Analyze_MinimizeSize.h
@@ -747,6 +747,7 @@ public :
     inline void Get_UTF16B (int64u Bytes, Ztring      &Info, const char*) {Get_UTF16B(Bytes, Info);}
     inline void Get_UTF16L (int64u Bytes, Ztring      &Info, const char*) {Get_UTF16L(Bytes, Info);}
     void Peek_Local (int64u Bytes, Ztring      &Info);
+    void Peek_UTF8  (int64u Bytes, Ztring      & Info);
     void Peek_String(int64u Bytes, std::string &Info);
     void Skip_Local (int64u Bytes,                    const char*) {if (Element_Offset+Bytes>Element_Size) {Trusted_IsNot(); return;} Element_Offset+=(size_t)Bytes;}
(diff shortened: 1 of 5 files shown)


cjee21:
Just realised the GPSAltitudeRef is not used. We need to read that to determine sign of GPSAltitude.

cjee21:
Tested with all files used to test previous Exif work. No regression found in text output.

@JeromeMartinez Is it expected that `Title` when mapped to `General_Title` is displayed as `Track name`?

JeromeMartinez:
"Orientation" is more complex topic because I already have a "Rotation" field in video tracks and I want to avoid similar fields, so I plan to reuse "Orientation" but I need to see how I manage H/V flips.
Please remove the Orientation part which will be addressed in a specific topic for not blocking this PR.

JeromeMartinez:
> For multicode strings, it seems everyone is using Unicode (16-bits) as encoding but Exif 3.0 specs now states that it is UTF-8...

A possibility is to try UTF-8 and if it fails (wrong code points) and/or too short (catching a zero), fallback to UTF16?

JeromeMartinez:
> How to check? It's at line 1577.

Sample file please, and I implement that.

cjee21:
> Sample file with UTF16 please.

[UserComment.zip](https://github.com/user-attachments/files/20550936/UserComment.zip)

Inserted with Exiftool.

Source/MediaInfo/File__Analyze_MinimizeSize.h
```diff
@@ -747,6 +747,7 @@ public :
     inline void Get_UTF16B (int64u Bytes, Ztring      &Info, const char*) {Get_UTF16B(Bytes, Info);}
     inline void Get_UTF16L (int64u Bytes, Ztring      &Info, const char*) {Get_UTF16L(Bytes, Info);}
     void Peek_Local (int64u Bytes, Ztring      &Info);
+    void Peek_UTF8(int64u Bytes, Ztring& Info);
```
cjee21, line 750: Spacing not aligned with the others

cjee21:
> > there is no code to handle the other types like signed rational yet but I guess that can be another PR when it is implemented (probably need to find some samples having those values first).
> 
> Right, I implement when I have samples for test of real use cases.

Sorry, I saw wrongly. Looks like all types used by Exif are already handled including signed rational so nothing to do. Signed long is listed as used by Exif in the specs but there appears to be no tags that use this type.

PR #2321, opened 2025-06-03
Exif: Initial Makernote parsing

cjee21:
For Apple, Nikon and Sony.

### DIFF
--- a/Source/MediaInfo/Tag/File_Exif.cpp
+++ b/Source/MediaInfo/Tag/File_Exif.cpp
@@ -26,6 +26,7 @@
     #include "MediaInfo/Image/File_Jpeg.h"
 #endif
 #include <cmath>
+#include <memory>
 //---------------------------------------------------------------------------
 
 namespace MediaInfoLib
@@ -941,8 +942,10 @@ enum kind_of_ifd
     Kind_Interop,
     Kind_MakernoteApple,
     Kind_MakernoteNikon,
+    Kind_NikonPreview,
     Kind_MakernoteSony,
     Kind_ParsingThumbnail,
+    Kind_ParsingNikonPreview
 };
 exif_tag_desc_size Exif_Descriptions[] =
 {
@@ -953,6 +956,7 @@ exif_tag_desc_size Exif_Descriptions[] =
     DESC_TABLE(IFDInterop, "Interoperability")
     DESC_TABLE(IFDMakernoteApple, "Apple Makernote")
     DESC_TABLE(IFDMakernoteNikon, "Nikon Makernote")
+    DESC_TABLE(IFD0, "Nikon Makernote Preview Image")
     DESC_TABLE(IFDMakernoteSony, "Sony Makernote")
 };
 static string Exif_Tag_Description(int8u NameSpace, int16u Tag_ID)
@@ -1848,6 +1852,13 @@ void File_Exif::Header_Parse()
         return;
     }
 
+    //Nikon Preview Image
+    if (currentIFD == Kind_ParsingNikonPreview) {
+        Header_Fill_Code(0xFFFFFFFF, "Nikon Makernote Preview Image");
+        Header_Fill_Size(Buffer_Size - Buffer_Offset);
+        return;
+    }
+
     //Get number of directories for this IFD
     int16u NrOfDirectories;
     Get_X2 (NrOfDirectories,                                    "NrOfDirectories");
@@ -1862,7 +1873,7 @@ void File_Exif::Data_Parse()
 {
     if (IfdItems.empty())
     {
-        if (currentIFD == Kind_ParsingThumbnail) {
+        if (currentIFD == Kind_ParsingThumbnail || currentIFD == Kind_ParsingNikonPreview) {
             Thumbnail();
         }
         else {
@@ -1916,6 +1927,21 @@ void File_Exif::Data_Parse()
             }
         }
 
+        //Nikon Preview Image
+        if (currentIFD != Kind_ParsingNikonPreview) {
+            const auto Infos_Thumbnail_It = Infos.find(Kind_NikonPreview);
+            if (Infos_Thumbnail_It != Infos.end()) {
+                const auto& Infos_Thumbnail = Infos_Thumbnail_It->second;
+                const auto ImageOffset = Infos_Thumbnail.find(IFD0::ImageOffset);
+                if (ImageOffset != Infos_Thumbnail.end() && ImageOffset->second.size() == 1) {
+                    int32u IFD_Offset = ImageOffset->second.at(0).To_int32u();
+                    GoToOffset(IFD_Offset);
+                    currentIFD = Kind_ParsingNikonPreview;
+                    return;
+                }
+            }
+        }
+
         Finish(); //No more IFDs
     }
 }
@@ -1953,6 +1979,9 @@ void File_Exif::Read_Directory()
         }
         else if (currentIFD == Kind_Exif && IfdItem.Tag == IFDExif::InteroperabilityIFD) {
             Get_IFDOffset(Kind_Interop);
+        }
+        else if (currentIFD == Kind_MakernoteNikon && IfdItem.Tag == IFDMakernoteNikon::PreviewIFD) {
+            Get_IFDOffset(Kind_NikonPreview);
         }
         else {
             GetValueOffsetu(IfdItem);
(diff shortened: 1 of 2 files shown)


cjee21:
Added parsing and display of a preview image found in Nikon Makernote. Sony Makernote also contains a preview image but it is located at the end of the file so it seems there is problem seeking there from IFD section. Some files have 3 thumbnails/previews, one standard Exif thumbnail, one Makernote preview image and one MPF Large Thumbnail.

Source/MediaInfo/Tag/File_Exif.cpp
```diff
@@ -982,12 +1171,141 @@ static const char* Exif_IFDGPS_GPSAltitudeRef_Name(int8u value)
...
+    case 1: return "Super Fine";
+    case 2: return "Fine";
+    case 3: return "Standard";
+    case 4: return "Economy";
+    case 5: return "Extra Fine";
+    case 6: return "RAW + JPEG/HEIF";
+    case 7: return "Compressed RAW";
+    case 8: return "Compressed RAW + JPEG";
+    case 9: return "Light";
+    case 0xFFFFFFFF: return "n/a";
```
JeromeMartinez, line 1281:
```suggestion
```

No n/a in MediaInfo, just empty.

Source/MediaInfo/Tag/File_Exif.cpp
```diff
@@ -982,12 +1171,141 @@ static const char* Exif_IFDGPS_GPSAltitudeRef_Name(int8u value)
...
+    }
+}
+
 //***************************************************************************
 // Constructor/Destructor
 //***************************************************************************
 
 //---------------------------------------------------------------------------
-File_Exif::File_Exif() : LittleEndian{}, currentIFD(Kind_IFD0)
+File_Exif::File_Exif() : LittleEndian{}, currentIFD(Kind_IFD0), MakernoteOffset{}
```
JeromeMartinez, line 1308:
```suggestion
File_Exif::File_Exif() : LittleEndian{}, currentIFD(Kind_IFD0)
```

Source/MediaInfo/Tag/File_Exif.h
```diff
@@ -34,6 +34,9 @@ class File_Exif : public File__Analyze
 
     //In
     bool FromHeif = false;
+    bool IsMakernote = false;
+    bool LittleEndian;
+    int32u MakernoteOffset;
```
JeromeMartinez, line 39:
```suggestion
    int32u MakernoteOffset = {};
```

Source/MediaInfo/Tag/File_Exif.cpp
```diff
@@ -745,6 +746,185 @@ exif_tag_desc Desc[] =
 };
 };
 
+namespace IFDMakernoteApple {
+    ELEM(0x0001, MakerNoteVersion)
+    ELEM(0x0003, RunTime)
```
JeromeMartinez, line 751: Just a question there, is it because you have no file with e.g... 2 or 10 or is it because it si opaque to you?
If it is because no file, I have a file with tags not listed there.

Source/MediaInfo/Tag/File_Exif.cpp
```diff
@@ -1003,6 +1321,9 @@ void File_Exif::Streams_Finish()
     const auto Infos_Exif_It = Infos.find(Kind_Exif);
     const auto Infos_GPS_It = Infos.find(Kind_GPS);
     const auto Infos_Interop_It = Infos.find(Kind_Interop);
+    const auto Infos_MakernoteApple_It = Infos.find(Kind_MakernoteApple);
+    const auto Infos_MakernoteNikon_It = Infos.find(Kind_MakernoteNikon);
+    const auto Infos_MakernoteSony_It = Infos.find(Kind_MakernoteSony);
```
JeromeMartinez, line 1326: I want to change that for others in the future, for the new items please move these lines just before the "if" for less diff.

Source/MediaInfo/Tag/File_Exif.cpp
```diff
@@ -1352,6 +1673,56 @@ void File_Exif::Streams_Finish()
...
+    if (Infos_MakernoteNikon_It != Infos.end()) {
+        currentIFD = Kind_MakernoteNikon;
+        const auto& Infos_MakernoteNikon = Infos_MakernoteNikon_It->second;
+        for (const auto& Item : Infos_MakernoteNikon) {
+            size_t Parameter = 0;
+            Ztring Value;
+            const char* ParameterC = nullptr;
+            string ParameterS;
+            switch (Item.first) {
+            case IFDMakernoteNikon::ActiveDLighting: ParameterC = "Active D-Lighting"; Value = Exif_IFDMakernoteNikon_ActiveDLighting_Name(Item.second.Read().To_int16u()); break;
```
JeromeMartinez, line 1702: No space acceptable in Parameter names.

Source/MediaInfo/Tag/File_Exif.cpp
```diff
@@ -1374,6 +1745,48 @@ void File_Exif::FileHeader_Parse()
...
+            if (alignment == 0x4D4D)
+                LittleEndian = false;
+            else {
+                Reject();
+                return;
+            }
+            currentIFD = Kind_MakernoteApple;
+            Accept();
+            Stream_Prepare(Stream_Image);
+            return;
```
JeromeMartinez, line 1763: I don't like the idea of returning there, is it possible to avoid the "return" here?

Source/MediaInfo/Tag/File_Exif.cpp
```diff
@@ -1374,6 +1745,48 @@ void File_Exif::FileHeader_Parse()
...
+        }
+        else if (Element_Size >= 7 && !strncmp((const char*)Buffer + Buffer_Offset, "Nikon\0\2", 7)) {
+            Skip_String(7,                                      "Identifier");
+            Skip_XX(3,                                          "(unknown)");
```
JeromeMartinez, line 1767:
```suggestion
            Skip_XX(3,                                          "(Unknown)");
```

Source/MediaInfo/Tag/File_Exif.cpp
```diff
@@ -1603,11 +2045,24 @@ void File_Exif::MulticodeString(ZtringList& Info)
...
-    Fill(Stream_Image, 1, Image_Type, "Thumbnail");
-    Fill(Stream_Image, 1, Image_MuxingMode, "Exif");
 
-    File__Analyze* Parser = nullptr;
-    const auto Infos_Thumbnail_It = Infos.find(Kind_IFD1);
+    std::unique_ptr<File__Analyze> Parser;
+    kind_of_ifd IFD;
+
+    if (currentIFD == Kind_ParsingNikonPreview) {
+        Fill(Stream_Image, StreamPos_Last, Image_Type, "Preview Image");
```
JeromeMartinez, line 2053: I see no practical difference between "Preview Image" and "Thumbnail", I would put "Thumbnail" for all.

Source/MediaInfo/Tag/File_Exif.h
```diff
@@ -34,6 +34,9 @@ class File_Exif : public File__Analyze
 
     //In
     bool FromHeif = false;
+    bool IsMakernote = false;
+    bool LittleEndian;
```
JeromeMartinez, line 38:
```suggestion
```

Not needed to move it IIUC.

JeromeMartinez (commented):
Thank you a lot for this PR, very useful at long term (need more mapping but it is already a very good start).

> Sony Makernote also contains a preview image but it is located at the end of the file so it seems there is problem seeking there from IFD section.

If it points to outside of the Exif block, it should be sent to the JPEG parser as it is done with the MPF PR. Could be done after the MPF PR is merged.
Could you provide a sample file?

> There is a bug where the standard Exif thumbnail size shown by MediaInfo is larger than the actual size.

Sample file please.

> The remaining bug I noticed is this IFD size for Nikon:

Sample file please.
Generally speaking, invalid value like 0 for offset should be discarded.

JeromeMartinez (changes requested):
(See above)

Source/MediaInfo/Tag/File_Exif.cpp
cjee21: 0x0002 is uncertain. 0x0010 is not listed by Exiftool. For the larger numbers it will be either uncertain, not listed or I have not done until there for brands with long list of tags.

Source/MediaInfo/Tag/File_Exif.h
cjee21: Right, same class can access private variables regardless of instance.

cjee21:
> Sample file please.

For Nikon, here are two samples with Adobe RGB and sRGB both JPEG and RAW that can reproduce the two size bugs: [Nikon.zip](https://github.com/user-attachments/files/20639587/Nikon.zip)

For Sony, this link contains samples for one of the older models that have preview image in Makernote: https://www.dpreview.com/products/sony/slrs/sony_slta35/sample-photos

PR #2323, opened 2025-06-09
Fix IAMF parsing from URL, typos and inconsistencies

cjee21:
IAMF and AVS3V were not added to MultipleParsing although their headers have been included there.

Also fix some typos and inconsistencies that I noticed previously while working on other things.


cjee21:
@JeromeMartinez about this for the XMP elements:

```C++
Element_Show(); //TODO: why is it needed?
```

I think I found the cause. It is because the XMP parser does not output anything to trace at all and it causes this line that hides the element to execute:

https://github.com/MediaArea/MediaInfoLib/blob/e9b6d4c0a45ad47c2a36a3ac5590e860c0682960/Source/MediaInfo/File__Analyze.cpp#L929

JeromeMartinez:
> I think I found the cause. It is because the XMP parser does not output anything to trace at all and it causes this line that hides the element to execute:

Hmmm... I don't remind the reason I added that, maybe it was not a good reason and/or the impact of this line is too much. I'll remove it and check if it has unexpected behavior.

> By the way, why does File__MultipleParsing.h have the following header?

Stupid copy paste years ago, I guess...

> IAMF and AVS3V were not added to MultipleParsing although their headers have been included there.

IIRC it was on purpose, the parsers have too many false positive (detecting a stream with these formats even if it is not the case) and AFAIK there is no real raw content, is there an issue for someone not to have them there?

JeromeMartinez:
> I am not sure about the others but for IAMF and WebP, it might be useful to be able to open directly from URLs?

I would be fine with doing that if there is not too much false positives, else we need to improve the parsers for avoiding the false positives.
I'll run MI with both parsers in priority and we'll see how it goes.

PR #2324, opened 2025-06-09
HEVC: align HDR Vivid parsing with T/UWA 005.1 specification

paulhiggs:
update parsing of HDR Vivid SEI message in HEVC. 

1. align element names with those defined in [T/UWA 005.1](https://uhd-world-association.com/wp-content/uploads/2025/03/TUWA-005.1-2022-HDR-Video-Technology-Part-1-Metadata-and-Tone-Mapping.pdf). 
2. move `color_saturation_mapping` information into second `num_windows` loop
3. read 'boolean values' as integers which represent either 1 or 2


JeromeMartinez:
The PR introduces no regression but also provides no additional output for the files I have so I rely on format experts for the usefulness of this PR, but if possible a file showing the difference would be great.

PR #2325, opened 2025-06-10
JPEG/PNG/XMP: Handle more metadata

cjee21:


JeromeMartinez:
> For some reason, filling custom parameter names in XMP will not be displayed in text view if other parsers have filled data

I saw that there are some issues when some metadata comes from different places, I have a patch in the queue but it has currently some unexpected side effects, WIP.

BTW, it would be great if you can provide sample files for seeing the changes from this PR (AFK enjoying the Dutch train strike but I hope to be able to review all your great PRs next week after a couple of other urgent tasks I have to do before that).

JeromeMartinez:
> For the Google AI tags and Exif encoded as zTXt:

By curiosity, why the PNG have a .jpg extension?

cjee21:
> By curiosity, why the PNG have a .jpg extension?

Ask Google 😉 I did not modify it straight from Gemini app. I guess they forgot to change the extension after switching from Imagen model to the native image generator. I think Microsoft's AI tools have saved jpg with .png extension before.

JeromeMartinez:
I adapted the PR.

IPTC in JPEG: I disagree with the Photoshop parsing in JPEG parser, as "8BIM" is part of PSD specs, so I expanded the PSD parser for parsing "8BIM" then I modified your commit for binding to the PSD parser. At the same time I managed the thumbnails in PSD so in JPEG (yet another thumbnail!). You can expand it from there for parsing the IPTC, and/or adding more metadata (I see XMP in Photoshop part of JPEG, crazy...)

PNG: I prefer to avoid tracking copyrighted code for small parts, I rewritted the HexStringToBytes() function (also adding a boundary check, avoiding many call of strtol for 2 bytes and avoiding malloc then copy) and other parts are too much bound to the actual file content for copyrights. I also added the support of "Raw profile type exif". I also parse now the full PNG for catching Exif metadata at the end of some files I have.

@cjee21 please review.

cjee21:
I renamed the copied-pasted Exif codes to Psd and fixed the PSD thumbnail getting merged with the wrong image streams.

> thumbnails in PSD so in JPEG (yet another thumbnail!)

The more we dig, the more thumbnails we find.... Exif then Makernotes then MPF now PSD... will there be a thumbnail in XMP too? 😅 

> I see XMP in Photoshop part of JPEG, crazy...

The more I see, the more crazy it looks. Some similar metadata getting stored in different places and various formats contained in each other.

There is one more thing.... Images with embedded videos: https://developer.android.com/media/platform/motion-photo-format
I do not have any such files.

cjee21:
By the way are the Extended XMP and C2PA parsers for JPEG properly deleted/freed after parsing?

cjee21:
Fixed non-MSVC compilation errors regarding `std::unique_ptr` and `std::move`.

Tried to parse some PSD metadata.

JeromeMartinez:
> PNG: Correct Exif stream merging

I definitely need to have a rule for this kind of stuff, maybe a dedicated method, but, well, for later :).

PR #2327, opened 2025-06-14
WavPack: Display MD5 sum of raw audio data

### LINKED ISSUE #2326: Display MD5 when analyzing WavPack file (similar to FLAC)
Windows 7 x64, Mediainfo 25.04.

Source: [siren.zip](https://github.com/user-attachments/files/20734748/siren.zip)

## FLAC

``` console
$ mediainfo siren.flac

General
Complete name                            : siren.flac
Format                                   : FLAC
Format/Info                              : Free Lossless Audio Codec
File size                                : 3.15 MiB
Duration                                 : 31 s 200 ms
Overall bit rate mode                    : Variable
Overall bit rate                         : 846 kb/s

Audio
Format                                   : FLAC
Format/Info                              : Free Lossless Audio Codec
Duration                                 : 31 s 200 ms
Bit rate mode                            : Variable
Bit rate                                 : 844 kb/s
Channel(s)                               : 2 channels
Channel layout                           : L R
Sampling rate                            : 44.1 kHz
Compression mode                         : Lossless
Stream size                              : 3.14 MiB (100%)
Writing library                          : libFLAC 1.5.0 (2025-02-11)
MD5 of the unencoded content             : 537832DCEB2CC889028721A8D718F0E2
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                                           this! this! this!
```

## WavPack

``` console
$ mediainfo siren.wv

General
Complete name                            : siren.wv
Format                                   : WavPack
File size                                : 3.14 MiB
Duration                                 : 31 s 200 ms
Overall bit rate mode                    : Variable
Overall bit rate                         : 845 kb/s

Audio
Format                                   : WavPack
Format profile                           : 4.16
Duration                                 : 31 s 200 ms
Bit rate mode                            : Variable
Bit rate                                 : 845 kb/s
Channel(s)                               : 2 channels
Sampling rate                            : 44.1 kHz
Compression mode                         : Lossless
Compression ratio                        : 1.669
Stream size                              : 3.14 MiB (100%)
Encoding settings                        : -x3 -m
```

## Wish

Dear @JeromeMartinez, could you be so kind as to add MD5 field to WavPack analysis?

For example, Foobar2000 has been able to display it for a long time.

cc: @dbry

cjee21:
Resolves #2326

### DIFF
--- a/Source/MediaInfo/Audio/File_Wvpk.cpp
+++ b/Source/MediaInfo/Audio/File_Wvpk.cpp
@@ -85,11 +85,20 @@ static const char* Wvpk_id(int8u ID)
         case 0x0B : return "correction file bitstream (wvc file)";
         case 0x0C : return "special extended bitstream for floating point data or integers > 24 bit";
         case 0x0D : return "contains channel count and channel_mask";
+        // ids from here are "optional" so decoders should skip them if they don't understand them
         case 0x21 : return "RIFF header for .wav files (before audio)";
         case 0x22 : return "RIFF trailer for .wav files (after audio)";
         case 0x25 : return "some encoding details for info purposes";
         case 0x26 : return "16-byte MD5 sum of raw audio data";
         case 0x27 : return "non-standard sampling rate info";
+        // added with version 5.0 to handle non-wav files and block checksums
+        case 0x23 : return "header for non-wav files";
+        case 0x24 : return "trailer for non-wav files";
+        case 0x28 : return "target filename extension";
+        case 0x29 : return "16-byte MD5 sum of raw audio data with non-wav standard";
+        case 0x2a : return "new file configuration stuff";
+        case 0x2b : return "identities of non-MS channels";
+        case 0x2f : return "2- or 4-byte checksum of entire block";
         default:    return "";
     }
 }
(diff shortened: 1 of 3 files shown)


dbry:
Does this actually work for larger files?

One of the unfortunate thing about WavPack files is that the MD5 is stored in the _last_ frame of the file, just before the APEv2 tag (if present). This was done so WavPack files could be created without having to back up to write the MD5 early in the file, but it means that to read the MD5 you have to either parse the whole file, or seek to the end and find the last frame (which is not that easy). I do this in `libwavpack`, but I don't see that here.

Thanks!

JeromeMartinez:
> The 512KiB from the end is from the beginning of the ApeTag it appears.

It is :-p.

dbry:
> Tested with file that has large cover art in APEv2 tag and it works too. The 512KiB from the end is from the beginning of the ApeTag it appears.

Oops, sorry, I missed that when I was looking through the code!

The only thing I would suggest is that 1 MB would be a safer value because _in theory_ WavPack frames can be as large as 1 MB. You might want to do that only if no frame is discovered at 512KiB, which would be faster for most files. Of course, what you have now will work for most files.

Thanks again for the great WavPack support!

dbry:
@cjee21 I just noticed something else. There are two metadata IDs for MD5 sums. The one you used was 0x26 which is the general case. However, for non-wav files that might have special requirements for calculating the MD5 sum (e.g., big-endian audio from AIFF files) there's another ID code (0x29). This was done to prevent older software (including wvunpack) from falsely triggering a verification error.

You mention these in the code, but are not handling the newer case. The handling is identical if you're just displaying the sum, so the easiest fix would be to add 0x29 to the same switch case as 0x26. You might want to rename `File_Wvpk::id_26` to `File_Wvpk::id_md5` also, just to be clear.

Cheers!

PR #2331, opened 2025-06-18
WavPack: Display MD5 sum of raw audio data (DSD)

cjee21:
Fix https://github.com/MediaArea/MediaInfoLib/pull/2327#issuecomment-2981758958

### DIFF
--- a/Source/MediaInfo/Audio/File_Wvpk.cpp
+++ b/Source/MediaInfo/Audio/File_Wvpk.cpp
@@ -6,7 +6,8 @@
 
 //+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 //
-// Source : http://www.wavpack.com/file_format.txt
+// Sources : http://www.wavpack.com/file_format.txt
+//           https://www.wavpack.com/WavPack5FileFormat.pdf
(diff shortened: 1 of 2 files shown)


Source/MediaInfo/Audio/File_Wvpk.cpp
```diff
@@ -112,7 +114,7 @@ extern std::string ExtensibleWave_ChannelMask_ChannelLayout(int32u ChannelMask);
 
 //---------------------------------------------------------------------------
 File_Wvpk::File_Wvpk()
-:File__Analyze(), File__Tags_Helper()
+:File__Analyze(), File__Tags_Helper(), Size{}, block_samples_LastFrame{}, cross_channel_decorrelation{}, joint_stereo{}, version{}
```
JeromeMartinez, line 117: I am moving to set initialization in the class definition (.h), please adapt the patch for that if you want to initialize everything.

cjee21: I just fixed the warnings on the files that I am working on to reduce the overall warnings. I'll just remove it from this and leave it to when someone wants to fix all the warnings.

PR #2335, opened 2025-06-18
Parse DNG & NEF

cjee21:


cjee21:
Current status:

Found full-sized JPEG preview and Nikon Compressed RAW image in SubIFDs for NEF.
Added temporary commit to confirm existence of JPEG at expected position in file.
To see how to show those in output using a method that will work for all Tiff-based files.

Known issues:

`Format` is not shown in MediaInfo output although it is filled for NEF. `Format/Info` shows without issues.

JeromeMartinez:
Thanks.
Keep the commit, I want to see every image in the files.

I'll check for "Format".

JeromeMartinez:
>   I just add support for (a randomly chosen number) 3 SubIFDs. 

I am fine with that, no need to manage non existant files, we'll see at long term...

cjee21:
There is an issue with Compression 6 and 7 JPEG. The names are different in Tiff and Exif codes. The handling is also wrong for Thumbnail parsing. The 7 one is stored as strips and can be found in Sony ARW. Sony ARW seems complex (where is the actual RAW data?) but the first image is a JPEG thumbnail which is not parsed.

Probably need to refactor the Tiff.cpp to loop through main and sub IFDs, detect the compression types and parse/fill accordingly.

JeromeMartinez:
> The handling may also be wrong for Thumbnail parsing (can JPEG parser handle the strips?). 

We don't parse this part so it should ne be impacting.

>  Sony ARW seems complex (where is the actual RAW data?) but the first image is a JPEG thumbnail (type 6) which is not parsed.

Not parsed? Isn't it just TIFF compatible so displayed as the first image? Sample file?

> Probably need to refactor the Tiff.cpp to loop through main and sub IFDs, detect the compression types and parse/fill accordingly.

If too difficult for you, let me know ( with a file if it is specific) and I look at it.

cjee21:
> If you have a link or attach here :-p.

Sony ARW: https://www.dpreview.com/sample-galleries/5968723450/sony-zv-e10-ii-sample-gallery/9709838505

Nikon NRW: https://www.dpreview.com/sample-galleries/8732294445/nikon-coolpix-p950-sample-gallery/7330085905

JeromeMartinez:
> Will it be a good idea to handle the SubIFDs in the same way as IFD0 in Tiff.cpp? 

It may be worth a try.

JeromeMartinez:
@cjee21 in practice I rely a lot on your tentatives and I adapt the code when you are blocked, so if you think that it is a good option, please try and then I adapt if it is needed. What is important to me first is to have a proof of concept then I adapt for more sustainability if needed.

cjee21:
> It may be worth a try.

I shift the SubIFDs enum next to IFD0 for now, maybe can iterate.

cjee21:
ARW has IFD2 so added that. There appears to be a crazy amount of private IFDs in ARW. The Makernote offset in ARW is also pointing to after the Makernote header identifier so the current codes does not detect it.

cjee21:
Added XMP, PS and IPTC parsing as well as added a few types and tags. XMP and IPTC is tested but Photoshop not tested as I have no Photoshop-generated TIFFs.

cjee21:
> Sony ARW seems complex (where is the actual RAW data?)
 
Looks like the large type 7 JPEG (ISO) compression one is the actual RAW image. It is compressed with Lossless JPEG (1992) ([source](https://github.com/dnglab/dnglab/blob/1a9191cb7803d469dcbdf191c7da4f24783179c0/rawler/src/decoders/arw.rs#L131)). Different model use different compression types, total 3 types for Sony.

JeromeMartinez:
> Tiff/Exif: Update compression values
Tiff/Exif: JPEG tags from specification
Exif: IFD data type
Tiff/Exif: Parse XMP
Tiff/Exif: Parse PSIR
Tiff/Exif: Parse IPTC-NAA

theses commits are easy to review and not impacting a lot the parsing of other files, please send them in a separate PR.

cjee21:
> theses commits are easy to review and not impacting a lot the parsing of other files, please send them in a separate PR.

Sent in 2347 and now this PR is built on that one.

cjee21:
This PR should be able to display all images in most if not all DNG/Sony/Nikon RAW files already. The only thing remaining is if you want to parse the JPEG images using JPEG parser since some of the JPEG images so not have info in the IFD so some of the streams are quite empty without even image dimensions.

Canon RAW files are not handled. Canon has 3 generations of RAW files. One using proprietary format, one Tiff based and the newest is ISOBMFF / MP4 based.

Source/Resource/Text/DataBase/Format.csv
```diff
@@ -125,9 +125,11 @@ Wave;;;A;Riff;;act at9 wav;audio/vnd.wave;;
 Wave64;;;A;Riff;;w64;;;
 WavPack;;;A;Wvpk;;wv wvc;;http://www.wavpack.com;
 Arri Raw;;;I;ArriRaw;;ari;;;
+ARW;;;I;Tiff;Sony Alpha RAW;arw;image/x-sony-arw;http://www.sony.com/;
```
JeromeMartinez, line 128: They are names with company name in the other PR about TIFF compression formats, so for coherency it should be there too, I think.

Source/MediaInfo/Image/File_Tiff.cpp
```diff
@@ -204,141 +204,165 @@ static const char* Tiff_ExtraSamples_ColorSpace(int32u ExtraSamples)
 void File_Tiff::Streams_Finish()
 {
     if (Retrieve_Const(Stream_General, 0, General_Format).empty()) {
-        Fill(Stream_General, 0, General_Format, "TIFF");
+        const auto FileExt{ Retrieve_Const(Stream_General, 0, General_FileExtension) };
+        if (FileExt == __T("arw") || FileExt == __T("ARW"))
```
JeromeMartinez, line 208: The format should not depend on the file name (sometimes we don't have it), isn't a way in the tags to catch the format?

cjee21:
Removed the following from Exif.cpp as it causes Format to disappear and also conflicts with new type handling in Tiff.cpp.

```c++
case IFD0::SubfileType: {
    if (Item.second.Read().To_int64u() & 1) {
        Fill(Stream_Image, 0, Image_Type, "Thumbnail");
        Clear(Stream_General, 0, General_Format);
    }
    Parameter = (size_t)-1;
    break;
}
```

Nikon NEF filetype is now detected based on content. Still no idea how to detect ARW and NRW from content.

cjee21:
Updated NRW detection. Should be reliable enough now.

JeromeMartinez:
In order to avoid to let this PR open a too long time, I did some QA and added a commit for handling issues (not always related to this PR, just that I remarked the issues) I found + I reverted the TIFF "flavors" filling (NRW, NEF...) because it does not always match with the extensions of the files I have, so I prefer to keep "TIFF" until we find a better rule.

PR #2339, opened 2025-06-23
Parse gain map metadata

cjee21:
Parse ISO 21496-1 Gain map metadata for JPEG and PNG.

### DIFF
--- a/Source/MediaInfo/Image/File_Png.cpp
+++ b/Source/MediaInfo/Image/File_Png.cpp
@@ -30,6 +30,7 @@
 
 //---------------------------------------------------------------------------
 #include "MediaInfo/Image/File_Png.h"
+#include "MediaInfo/Image/File_GainMap.h"
 #include "MediaInfo/MediaInfo_Config_MediaInfo.h"
 #if defined(MEDIAINFO_JPEG_YES)
     #include "MediaInfo/Image/File_Jpeg.h"
@@ -123,6 +124,8 @@ namespace Elements
     const int32u fcTL=0x6663544C;
     const int32u fdAT=0x66644154;
     const int32u gAMA=0x67414D41;
+    const int32u gdAT=0x67644154;
+    const int32u gmAP=0x676D4150;
     const int32u hIST=0x68495354;
     const int32u iCCP=0x69434350;
     const int32u iTXt=0x69545874;
@@ -316,6 +319,8 @@ void File_Png::Data_Parse()
         CASE_INFO(fcTL,                                         "Frame Control Chunk"); 
         CASE_INFO(fdAT,                                         "Frame Data Chunk"); 
         CASE_INFO(gAMA,                                         "Gamma");
+        CASE_INFO(gdAT,                                         "Gain map image data");
+        CASE_INFO(gmAP,                                         "Gain map metadata");
         CASE_INFO(hIST,                                         "Image histogram"); 
         CASE_INFO(iCCP,                                         "Embedded ICC profile");
         CASE_INFO(iTXt,                                         "International textual data");
@@ -583,6 +588,36 @@ void File_Png::gAMA()
     FILLING_END()
 
     Data_Common();
+}
+
+//---------------------------------------------------------------------------
+void File_Png::gdAT()
+{
+    Stream_Prepare(Stream_Image);
+    Fill(Stream_Image, StreamPos_Last, Image_Type, "Gain map");
+
+    //Parsing
+    File_Png MI;
+    Open_Buffer_Init(&MI);
+    Open_Buffer_Continue(&MI);
+    Open_Buffer_Finalize(&MI);
+    Merge(MI, Stream_Image, 0, StreamPos_Last);
+    if (MI.GainMap_metadata_ISO) {
+        GainMap_metadata_ISO.reset(new GainMap_metadata());
+        *static_cast<GainMap_metadata*>(GainMap_metadata_ISO.get()) = *static_cast<GainMap_metadata*>(MI.GainMap_metadata_ISO.get());
+    }
+}
+
+//---------------------------------------------------------------------------
+void File_Png::gmAP()
+{
+    //Parsing
+    File_GainMap MI;
+    GainMap_metadata_ISO.reset(new GainMap_metadata());
+    MI.output = static_cast<GainMap_metadata*>(GainMap_metadata_ISO.get());
+    Open_Buffer_Init(&MI);
+    Open_Buffer_Continue(&MI);
+    Open_Buffer_Finalize(&MI);
 }
 
 //---------------------------------------------------------------------------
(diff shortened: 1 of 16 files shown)


cjee21:
> Sample file please :).

> Same :)

The same link as previously posted in Ultra HDR issue:

https://www.adobe.com/go/gainmap_sample_photos

For AVIF, the Exif will only parse after the fix in the Exif PR. The XMP is not handled/parsed. The ISO 21496-1 data will be shown in trace on this PR.

cjee21:
> No crash on my side
01_base_hdr.avif

Try 01_base_sdr.avif

JeromeMartinez:
> Try 01_base_sdr.avif

Same.

cjee21:
Looked at more samples and updated the parsing, edited above comment that is wrong. Should now be able to parse all available samples. Also found that for AVIF, it may be stored either in idat or mdat, so two temporary commits to see it for both cases.

JeromeMartinez:
>  In some cases it does not reach the Finish() point

Name or sample file please.

cjee21:
> > In some cases it does not reach the Finish() point
> 
> Name or sample file please.

9931408803 from Google Pixel 9 Pro. It seems there is early return somewhere when there is a `<?xpacket end="w"?>` or a self-closing `rdf:Description`.

JeromeMartinez:
> XMP parsing should be okay now. I modified the macros some more that were originally based on ones in spherical video.

The spherical video macros were based on the ADM macros, and I hope that at long term I could have identical XML parsing related code, in the meantime it is good to have such implementation so when I finalize common XML parsing code I have enough for testing.

> Let me know if you want to move the XMP fixes to a new PR.

It will be slow on my side for gain map, so if you need the XMP fixes for something else, I am fine for a dedicated PR and I merge it after quick regression checks, else it may stay there until I have the time for checking the whole PR.

JeromeMartinez (changes requested):
After a quick review, the quality of the PR seems fine (also AVIF) for merging it, but it seems that I don't have enough test files, I don't find something with the XMP stuff or the PNG, could you provide sample files?

And just a minor change request.

cjee21:
> It will be slow on my side for gain map, so if you need the XMP fixes for something else, I am fine for a dedicated PR and I merge it after quick regression checks

I don't think I need it for something else but since the XMP PR is already merged, I think it'll be good to get it fixed earlier in case there are files that don't parse properly so I made the new PR.

> After a quick review, the quality of the PR seems fine (also AVIF) for merging it

The AVIF part (the TEMP commits) are definitely not good. That is a really quick and 'hackish' way that would not work properly for different files. It likely requires some restructuring to handle the different construction methods which I do not have an idea how.

PR 2334 should be easy to review and maybe also 2336. The others require more work (likely need you to do some restructuring) so those can be put to the following release after a new release version if there is a release soon.

JeromeMartinez:
> The AVIF part (the TEMP commits) are definitely not good. That is a really quick and 'hackish' way that would not work properly for different files. It likely requires some restructuring to handle the different construction methods which I do not have an idea how.

It seems to work well on files I tested, please an example of non working file.

JeromeMartinez:
> One of them parses gain map from Adobe samples. The other parses gain map from libavif test files. The 2nd TEMP comments out the first.

Got it, I'll test both and I'll find a way to merge them.

> >    or the PNG
> I do not have samples for this yet. Need a screenshot of HDR content from an Android 16 device.

I was thinking to merge this PR with PNG only but as we don't have file it can wait.

cjee21:
> Got it, I'll test both and I'll find a way to merge them.

For the IDAT, probably need to implement parsing similar to MDAT (the part I do not know how). So that can parse other metadata that may be in IDAT as well. Need to read the construction method and handle appropriately, whether it is mdat or idat.

> I was thinking to merge this PR with PNG only but as we don't have file it can wait.

If you want, can merge the JPEG part first (this PR has JPEG, PNG and AVIF but AVIF not ready and PNG no samples). But right now it is just parsing and populating a struct. Not yet do anything with the data. Probably in JPEG, check that we have gain map and metadata present (which type) then fill the HDR format along with necessary data.

JeromeMartinez:
>  then fill the HDR format along with necessary data.

For the moment it is not clear about how I would fill the HDR field, there are so many formats and hacks, I'll think about that but later.

cjee21:
Completed the PNG one. Now left the AVIF idat/mdat handling and filling HDR format.

cjee21:
Rebased + some simplification of XMP part.

cjee21:
What is blocking the AVIF gain map parsing is the `iloc` construction method.

https://github.com/MediaArea/MediaInfoLib/blob/b5e97926de3409bec2613f6673fbb723a4909494/Source/MediaInfo/Multiple/File_Mpeg4_Elements.cpp#L2649

Screenshot 2025-07-27 010455

The ISO 21496-1 data can be in either `mdat` or `idat` based on different sample files and they have different construction method.

I do not know how the current `mdat` parsing works. Seems complex. Once the construction method is handled, adding gain map parsing should be as simple as adding Exif parsing.

JeromeMartinez:
Please 2 files with different method, and I check.

JeromeMartinez:
>  It is not done properly (will not handle other data in idat or multiple extents, may also have code correctness issues).

It seems fine, at least with your version, do you still feel not confortable enough with it for not removing the "TEMP" thing?

JeromeMartinez (approved):

Source/MediaInfo/Image/File_Png.cpp
```diff
@@ -585,6 +590,34 @@ void File_Png::gAMA()
...
+    Fill(Stream_Image, StreamPos_Last, Image_Type, "Gain map");
+
+    //Parsing
+    File_Png MI;
+    Open_Buffer_Init(&MI);
+    Open_Buffer_Continue(&MI);
+    Open_Buffer_Finalize(&MI);
+    Merge(MI, Stream_Image, 0, StreamPos_Last);
+    GainMap_metadata_ISO.reset(new GainMap_metadata());
+    *static_cast<GainMap_metadata*>(GainMap_metadata_ISO.get()) = *static_cast<GainMap_metadata*>(MI.GainMap_metadata_ISO.get());
```
JeromeMartinez, line 606:
```suggestion
    if (MI.GainMap_metadata_ISO.get()) {
        GainMap_metadata_ISO.reset(new GainMap_metadata());
        *static_cast<GainMap_metadata*>(GainMap_metadata_ISO.get()) = *static_cast<GainMap_metadata*>(MI.GainMap_metadata_ISO.get());
    }
```

else crash with https://skia.googlesource.com/skia/+/418c68ea5ccb/resources/images/gainmap_gdat_no_gmap.png

JeromeMartinez (changes requested):

JeromeMartinez (approved):

PR #2342, opened 2025-06-24
IAMF: Add check for TestParser

### LINKED ISSUE #2340: [v25.04] MediaInfo::Open hangs on a specific MXF file
OS: Windows 11
MediaInfo version: v25.04

I have found an MXF file that causes the [MediaInfo::Open](https://github.com/MediaArea/MediaInfoLib/blob/e1ab0639d69668de1636dce47585182d7655fbbf/Source/MediaInfo/MediaInfo.h#L62) method to hang.
The file: [file_causing_hang.mxf.zip](https://github.com/user-attachments/files/20871569/file_causing_hang.mxf.zip)

You can also observe the hang by running the command `MediaInfo.exe file_causing_hang.mxf`

It only reproduces on `v25.04`, it doesn't reproduce on older versions.

g-maxime:
I'm not sure if this is the best solution.

Fixes #2340

### DIFF
--- a/Source/MediaInfo/Audio/File_Iamf.cpp
+++ b/Source/MediaInfo/Audio/File_Iamf.cpp
@@ -180,6 +180,27 @@ void File_Iamf::Streams_Accept()
 }
 
 //***************************************************************************
+// Buffer - File header
+//***************************************************************************
+
+//---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


cjee21:
This breaks IAMF parsing though...

IAMF in MP4 gets rejected. Standalone IAMF does not show Format profile.

PR #2344, opened 2025-06-27
XMP: Migrate to tfsxml

cjee21:
Since XMP parser codes are getting longer and more complicated, I tried migrating to tfsxml to see how it goes. Not thoroughly tested.


cjee21:
This actually works better than master branch for this PDF sample: https://tcpdf.org/examples/example_065/

There is no additional processing that was done by `FileHeader_Begin_XML`. I don't have files that need that.

JeromeMartinez:
Thank you for the migration. At long term I wish to migrate all XML parsers to tfsxml.
I am not fully happy with this PR because it demonstrates issues when there are conflicts between different sources and also there are issue when dates are a bit weird, but this can be for later.

PR #2348, opened 2025-07-02
Curate application/library/hardware name/model/version

JeromeMartinez:
A beginning, and not the best code, but it does the job


JeromeMartinez:
GIMP has hint in a lot of places and with different details, unfortunately there are issues in MediaInfo during merge of information from different parsers, I need to fix that first before handling completely GIMP.

cjee21:
Xperia Z3 (https://www.dpreview.com/sample-galleries/1160745414/sony-xperia-z3-review-samples) 'Writing application' is displayed differently compared to other Xperia. MediaInfo does not put hardware name there.

JeromeMartinez:
> Xperia Z3 (https://www.dpreview.com/sample-galleries/1160745414/sony-xperia-z3-review-samples) 'Writing application' is displayed differently compared to other Xperia. MediaInfo does not put hardware name there.

The version number is awful, not detected as an actual version number, but I'll figure out how to flag that as a version number.

PR #2351, opened 2025-07-04
Map Samsung Galaxy and Xiaomi model names

cjee21:


JeromeMartinez:
> Looks like some Samsung models have model names instead of model numbers in the metadata:

It is complicated, yes :).
And worse: `S918BXXU1AWA6` is a version number (the firmware version number).
From that, we can extract the model name (and the "application" version), but we need to figure out the format of the firmware number and how we apply it...

cjee21:
New Samsung samples: https://www.gsmarena.com/samsung_galaxy_z_fold7_and_z_flip7_camera_samples-news-68557.php

cjee21:
I'll check tomorrow.

> added space between Flip and the number

The official name does not have space that's why I removed it.

JeromeMartinez:
>  added space between Flip and the number
The official name does not have space that's why I removed it.

Very fun, it depends if you check the web page title or the web page content!
image

https://en.wikipedia.org/wiki/Samsung_Galaxy_Z_Fold_7
"The Samsung Galaxy Z Fold 7 (stylized as Samsung Galaxy Z Fold7)"

Lot of website add a space but as e.g  https://www.gsmarena.com/samsung_galaxy_z_fold7-13826.php does not put it, ok, no space, i update the commit for reverting my change.

cjee21:
> depends if you check the web page title or the web page content

I see that the name written by the phone in the metadata (if there is) has no space so I follow gsmarena.

PR #2353, opened 2025-07-06
MP4/MOV/HEVC: Ambient viewing environment

cjee21:
Parse Ambient viewing environment `amve` box.

> ```
> class AmbientViewingEnvironmentBox extends Box('amve') { 
>     unsigned int(32) ambient_illuminance;
>     unsigned int(16) ambient_light_x;
>     unsigned int(16) ambient_light_y;
> }
> ```

> **ambient_illuminance** specifies the environmental illluminance of the ambient viewing environment in units of 0.0001
lux. ambient_illuminance shall not be equal to 0.

> **ambient_light_x** and **ambient_light_y** specify the normalized x and y chromaticity coordinates, respectively, of the
environmental ambient light in the nominal viewing environment, according to the CIE 1931 definition of x and y as
specified in ISO 11664-1 (see also ISO 11664-3 and CIE 15), in normalized increments of 0.00002. The values of
ambient_light_x and ambient_light_y shall be in the range of 0 to 50 000, inclusive.
NOTE – For example, the conditions identified in Rec. ITU-R BT.2035 can be expressed using ambient_illuminance equal to 100 000 with background chromaticity indicating D65 (ambient_light_x equal to 15 635, ambient_light_y equal to 16 450), or optionally in some regions, background chromaticity indicating D93 (ambient_light_x equal to 14 155, ambient_light_y equal to 14 855)

Reference:
https://developer.apple.com/documentation/technotes/tn3145-hdr-video-metadata#Dolby-Vision-Profile-84
https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-H.265-202407-I!!PDF-E&type=items

Sample:
https://img.photographyblog.com/reviews/apple_iphone_13_pro/sample_images/4K60p.mov

### DIFF
--- a/Source/MediaInfo/File__Analyze.h
+++ b/Source/MediaInfo/File__Analyze.h
@@ -915,13 +915,16 @@ public :
     void Get_MasteringDisplayColorVolume(Ztring& MasteringDisplay_ColorPrimaries, Ztring& MasteringDisplay_Luminance, bool FromAV1=false);
     void Get_LightLevel(Ztring &MaxCLL, Ztring &MaxFALL, int32u Divisor=1);
     #endif
-    #if defined(MEDIAINFO_AV1_YES) || defined(MEDIAINFO_AVC_YES) || defined(MEDIAINFO_HEVC_YES) || defined(MEDIAINFO_MPEG4_YES) || defined(MEDIAINFO_MATROSKA_YES) || defined(MEDIAINFO_MXF_YES) || defined(MEDIAINFO_MPEGTS_YES)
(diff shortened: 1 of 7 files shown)


cjee21:
I do not know if it is desirable to display this information in the output.

JeromeMartinez:
> I do not know if it is desirable to display this information in the output.

It is.

JeromeMartinez:
@cjee21 see `MasteringDisplayColorVolume_Values_Compute` and the `/50000` stuff, it seems that something similar could be displayed for this information.

cjee21:
> @cjee21 see `MasteringDisplayColorVolume_Values_Compute` and the `/50000` stuff, it seems that something similar could be displayed for this information.

I found that, trying to understand it.

I see a bug there... `defined(MEDIAINFO_MATROSKA_YES)` should be `defined(MEDIAINFO_MK_YES)` ?

PR #2356, opened 2025-07-08
XMP: Simplify macros + Map all IPTC digital source types

cjee21:
Since the merging issue is now fixed, moved digital source type from copyright to its own field and map all types.

### DIFF
--- a/Source/MediaInfo/Tag/File_Xmp.cpp
+++ b/Source/MediaInfo/Tag/File_Xmp.cpp
@@ -445,14 +445,44 @@ void File_Xmp::xmp(const string& name, const string& value)
 //---------------------------------------------------------------------------
 void File_Xmp::Iptc4xmpExt(const string& name, const string& value)
 {
+    struct DigitalSourceType_mapping {
+        const char* URI;
+        const char* Name;
+    };
+    static const DigitalSourceType_mapping DigitalSourceTypes[] = {
+        { "http://cv.iptc.org/newscodes/digitalsourcetype/digitalCapture", "Digital capture sampled from real life" },
+        { "http://cv.iptc.org/newscodes/digitalsourcetype/computationalCapture", "Multi-frame computational capture sampled from real life" },
+        { "http://cv.iptc.org/newscodes/digitalsourcetype/negativeFilm", "Digitised from a transparent negative" },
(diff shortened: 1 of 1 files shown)


Source/MediaInfo/Tag/File_Xmp.cpp
```diff
@@ -468,14 +445,39 @@ void File_Xmp::xmp(const string& name, const string& value)
...
         string URI{ value };
         auto pos = URI.find("https://");
         if (pos != string::npos) URI.replace(pos, 5, "http"); // Some Google generated files have https instead of http
-        if (URI == "http://cv.iptc.org/newscodes/digitalsourcetype/trainedAlgorithmicMedia")
-            Fill(Stream_General, 0, General_Copyright, "Created using generative AI");
-        if (URI == "http://cv.iptc.org/newscodes/digitalsourcetype/compositeWithTrainedAlgorithmicMedia")
-            Fill(Stream_General, 0, General_Copyright, "Edited using generative AI");
+        for (int8u i = 0; i < DigitalSourceTypes_size; ++i) {
+            if (URI == DigitalSourceTypes[i].URI)
+                Fill(Stream_General, 0, "DigitalSourceType", DigitalSourceTypes[i].Name);
```
JeromeMartinez, line 479:
```suggestion
        if (pos != string::npos) {
            URI.erase(9, 1); // Some Google generated files have https instead of http
        }
        Fill(Stream_General, 0, "DigitalSourceType", URI);
        Fill_SetOptions(Stream_General, 0, "DigitalSourceType", "N NTY");
        int8u i = 0;
        for (; i < DigitalSourceTypes_size; ++i) {
            if (URI == DigitalSourceTypes[i].URI) {
                break;
            }
        }
        Fill(Stream_General, 0, "DigitalSourceType/String", i < DigitalSourceTypes_size ? DigitalSourceTypes[i].Name : URI.c_str());
        Fill_SetOptions(Stream_General, 0, "DigitalSourceType/String", "Y NTN");
```

(replace also the https stuff, no need to use "replace" here)
I actually prefer to keep the URI because it is the standard, and also your code (also in the past) does not show when the URI is not known, so now I fill the URI in the report (hidden by default) and show the updated string if the URI is known else I show the URI directly.

JeromeMartinez (approved):

PR #2362, opened 2025-07-23
More DTS:X parsing.

Aleksoid1978:
Sample file:
[Trinnov_Trailer_Experience_DTSX-Pro.zip](https://github.com/user-attachments/files/21396344/Trinnov_Trailer_Experience_DTSX-Pro.zip)

If necessary, I can upload a full video clip with this audio track.


JeromeMartinez:
Thanks. Still not convinced about how we detect that but as we don't have any spec, we keep doing like that...

PR #2367, opened 2025-07-26
HDR10+: Fix maxscl values + add AV1 support

cjee21:
Needs more tests. Codes for AV1 based on codes from HEVC (with minor changes to fix compiler errors/warnings, no idea why what works in HEVC does not work in AV1). I do not have sufficient HDR10+ files to test.

### DIFF
--- a/Source/MediaInfo/Video/File_Av1.cpp
+++ b/Source/MediaInfo/Video/File_Av1.cpp
@@ -70,6 +70,20 @@ const char* Av1_seq_profile(int8u seq_profile)
 }
 
 //---------------------------------------------------------------------------
+const char* Av1_metadata_type(int8u metadata_type)
+{
+    switch (metadata_type)
+    {
+    case  0x1: return "METADATA_TYPE_HDR_CLL";
+    case  0x2: return "METADATA_TYPE_HDR_MDCV";
+    case  0x3: return "METADATA_TYPE_SCALABILITY";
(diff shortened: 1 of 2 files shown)


Source/MediaInfo/File__Analyze_Streams.cpp
```diff
@@ -984,6 +984,117 @@ void File__Analyze::Get_LightLevel(Ztring &MaxCLL, Ztring &MaxFALL, int32u Divis
 }
 #endif
 
+//---------------------------------------------------------------------------
+#if defined(MEDIAINFO_AV1_YES) || defined(MEDIAINFO_HEVC_YES) || defined(MEDIAINFO_MK_YES)
+void File__Analyze::Get_SMPTE_ST_2094_40(int8u& application_version, bool& IsHDRplus, bool& tone_mapping_flag)
+{
+    Get_B1 (application_version,                                "application_version");
+    if (application_version==1)
```
JeromeMartinez, line 992:
```suggestion
    if (application_version<=1)

```

JeromeMartinez (changes requested):
Thank you, good to have a common place for that.
The whole ancillary data would deserve a common parser but it is for later.

All is good except that version 0 is also supported in practice and it was in the MKV code, so use the MKV code for that (just a small change on the version check, if I remind correctly there is not so much difference between the versions)

PR #2365, opened 2025-07-25
Curate application/library/hardware name/model/version, update

JeromeMartinez:
Still lot of incoherence spot but for later.

PR #2368, opened 2025-07-26
Improve model / company name handling

cjee21:
Fix https://github.com/MediaArea/MediaInfoLib/pull/2365#issuecomment-3121464905 + code improvements.


JeromeMartinez:
This PR changes nearly (only Nokia "model" going in name, good) nothing in test suites from different sources 😅, they seem to be outdated, but I keep now the example you provided for not missing such case.

PR #2370, opened 2025-07-28
Build library for Android

g-maxime:
[Snapshot](https://mediaarea.net/download/snapshots/branches/binary/libmediainfo0/20250728_android-lib/MediaInfo_DLL_25.04.20250728_Android.zip)

### DIFF
--- a/Release/Release_DLL_Android.sh
+++ b/Release/Release_DLL_Android.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+##  Copyright (c) MediaArea.net SARL. All Rights Reserved.
+ #
+ #  Use of this source code is governed by a BSD-style license that can
+ #  be found in the License.html file in the root of the source tree.
+ ##
+
+#-----------------------------------------------------------------------
+# Helpers
+set -e # fail on any error
+releasedir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+#-----------------------------------------------------------------------
+# Used variables
+if [ -z "$VERSION" ] ; then
+    VERSION=$(<"$releasedir"/../Project/version.txt)
+fi
+
+if [ -z "$NDK" ] ; then
+    echo "error: missing mandatory environment \$NDK" >&2
+    exit 1
+fi
+
+if [ -z "$MINSDKVERSION" ] ; then
+    MINSDKVERSION=24 # Android 7.0 needed for -D_FILE_OFFSET_BITS=64
+fi
+
+#-----------------------------------------------------------------------
+# Clean
+rm -fr "$releasedir"/build
+rm -fr "$releasedir"/MediaInfo_DLL_Android
+rm -f "$releasedir"/MediaInfo_DLL_*_Android.zip
+
+#-----------------------------------------------------------------------
+# Build
+abis=(arm64-v8a armeabi-v7a x86_64 x86)
+for abi in "${abis[@]}" ; do
+    mkdir -p "$releasedir"/build/$abi
+    pushd "$releasedir"/build/$abi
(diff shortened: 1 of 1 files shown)


cjee21:
@g-maxime What is the intended use of this library? Looking at the script and binary size, it seems to be built in debug mode? Also it exports many symbols and does not appear to follow https://developer.android.com/ndk/guides/middleware-vendors regarding using the STL or Java middleware with JNI libraries.

cjee21:
> it seems to be built in debug mode?

Maybe it's built in release mode since there is DCMAKE_BUILD_TYPE=Release but it contains debug info that can be split.

Screenshot 2026-04-28 215449

```bash
llvm-objcopy --only-keep-debug libmediainfo.so libmediainfo.so.debug
llvm-objcopy --strip-debug libmediainfo.so      # maybe other command can strip even more stuff
llvm-objcopy --add-gnu-debuglink=libmediainfo.so.debug libmediainfo.so
```

```bash
llvm-readelf --dyn-symbols libmediainfo.so        # many symbols not needed to be exposed are visible
```

cjee21:
Using the new toolchain does not generate debug info in Release mode (RelWithDebInfo will have debug info). The legacy toolchain adds `-g` to the flags in both Release and RelWithDebInfo mode.

```
-DANDROID_USE_LEGACY_TOOLCHAIN_FILE=OFF
```

cjee21:
If it is intended to be used by others in Android app, I think can change from Release to RelWithDebInfo if intending to provide debug info (to make it clear that having debug info is intended). Then AGP should strip the binary when packaging a release APK.

For the other issue, can use version script to only make MediaInfo API symbols and JNI_OnLoad visible to reduce size of stripped binary and prevent symbol conflicts.

g-maxime:
Thank you.

The debuginfo is included on purpose to  allows third-party projects to receive relevant crash traces.; the binary is stripped during .aab generation.

I'll check for hiding unnecessary symbols.

cjee21:
> Using the new toolchain does not generate debug info in Release mode...

FYI the non-legacy toolchain is actually no longer supported.

> Warning: We do not recommend using -DANDROID_USE_LEGACY_TOOLCHAIN_FILE=OFF. The legacy toolchain is the default, and despite the name, is what we recommend.

> Earlier NDKs experimented with a new implementation of the toolchain file that would reduce behavior differences between using the NDK's toolchain file and using the built-in CMake support. This ended up requiring a significant amount of work (which has not been completed), but did not actually improve behavior, so we're no longer pursuing this.

cjee21:
Can also package as .aar for use with both Kotlin (via JNI wrapper) and CMake (via find_package using prefab). Example: https://github.com/cjee21/MediaInfoLib/tree/android-lib/Source/AndroidLib (generated AAR looks correct but did not test using it).

JeromeMartinez:
> Can also package as .aar for use with both Kotlin (via JNI wrapper) and CMake (via find_package using prefab)

I am totally newbe here, what it the advantage?
I am interested in everything "positive", but I prefer to understand it.

cjee21:
> I am totally newbe here, what it the advantage? I am interested in everything "positive", but I prefer to understand it.

I am new to this also. https://developer.android.com/studio/projects/android-library

Looks like this is more standard way and makes it easier for developers to use the library (just `implementation(files("my_path/my_lib.aar"))`) and also enabling prefab if wanting to use via native code with CMake (then can `find_package` in CMake and use as usual CMake library, header files included in AAR too). Also makes possible to [publish](https://developer.android.com/build/publish-library) to Maven for example if I understand correctly. It will also enforce the minSDK that we use to build the library so that consumers of it cannot set a lower minSDK then crashes the library at runtime.

cjee21:
> I am interested in everything "positive"

More positive things... For Windows, you can register [Windows Desktop Application Program](https://learn.microsoft.com/en-us/windows/win32/appxpkg/windows-desktop-application-program). Then if enable .pdb generation for MSVC release build and upload the .pdb file, you can retrieve and analyse crash dumps that Windows automatically collect from all users. Can also view how many users are using each Windows version. Then will know stability of the app in the wild and which versions of Windows are no longer used without relying only on user reports. (I never did it though, just from info I read.) PDB file in release mode also enables analysis with SizeBench to help optimize binary size.

FYI Generation of PDB in release mode was disabled in https://github.com/MediaArea/MediaInfoLib/commit/72248ebb693f3aeb527d5c3e61a7c60e8fe3104b

PR #2371, opened 2025-07-29
Preparing v25.07

g-maxime:


History_DLL.txt
```diff
@@ -7,6 +7,51 @@ x Correction
...
++ MPEG-4: Better support of Auro-Cx 3D audio (thanks to cjee21)
++ Matroska : Improvement of Spherical Video
++ Matroska: TimecodeXML output
++ AV1: Display of HDR10+ (thanks to cjee21)
++ CAF: Improved support (thanks to cjee21)
++ APE tags: Map more tags (thanks to cjee21)
++ WavPack: Display MD5 sum of raw audio data (thanks to cjee21)
++ FLAC: Map Arranger and Lyricist (thanks to cjee21)
++ DTS: More DTS:X detection
++ PNG/HEIF: Parse ISO 21496-1 Gain map metadata (thanks to cjee21)
```
cjee21, line 25: This should be JPG/PNG/AVIF?

cjee21: > don't think that it is enough for changing the release notes now

Yeah. The user visible change is gain map track for PNG and tmap track name for AVIF. The rest is only visible in trace for now.

JeromeMartinez: > The rest is only visible in trace for now.

We were not precise enough in the log about that, but not a big deal, people would ask and I plan to handle the display in the default report in the August release so we should have a release with a "good" support of it before most people remarks this lack of display in the default report.

cjee21: > before most people remarks this lack of display in the default report

Looks like the spec is just published earlier this month https://www.iso.org/standard/86775.html so should be okay.

PR #2372, opened 2025-07-29
AV1: Fix parsing

cjee21:
Parse `frame` OBU. Fix parsing `.obu` files and actually parse only the desired `Frame_Count_Valid` number of frames.

Ref: https://github.com/MediaArea/MediaInfo/issues/862#issuecomment-3132816445

### DIFF
--- a/Source/MediaInfo/Video/File_Av1.cpp
+++ b/Source/MediaInfo/Video/File_Av1.cpp
@@ -291,7 +291,7 @@ void File_Av1::Header_Parse()
 void File_Av1::Data_Parse()
 {
     //Probing mode in case of raw stream //TODO: better reject of bad files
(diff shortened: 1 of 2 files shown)


JeromeMartinez:
> SeenFrameHeader is only reset by temporal_delimiter but there is no temporal_delimiter when in MP4

I see it in the MP4 I tried but as it is optional, true we should not rely on it.

cjee21:
> I see it in the MP4 I tried but as it is optional, true we should not rely on it.

If I understand this correctly, it should not be present when in MP4.

https://aomediacodec.github.io/av1-isobmff/#sampleformat

"OBUs of type OBU_TEMPORAL_DELIMITER, OBU_PADDING, or OBU_REDUNDANT_FRAME_HEADER SHOULD NOT be used."

PR #2379, opened 2025-08-19
TrueHD: Improve support

cjee21:
More parsing and improved trace output for Dolby TrueHD. I put some of the new codes under a TrueHD/FBA check since I do not have MLP/FBB info or samples to know if it is the same. Maybe some of the dialnorm info should be shown in MediaInfo output. At the moment AC3/EAC3 have extra info in output but TureHD does not.


cjee21:
Tried to improve the `Format` `Format/Info` and `Commercial name` display but there is still some weirdness especially with `legacystreamdisplay` enabled:

```
Format                                   : MLP FBA / AC-3 AC-3 MLP FBA 16-ch
Format/Info                              : Audio Coding 3 + Meridian Lossless Packing FBA with 16-channel presentation
Commercial name                          : Dolby TrueHD with Dolby Atmos
Muxing mode                              : Stream extension
```

cjee21:
https://github.com/MediaArea/MediaInfoLib/blob/d93cc8c86260ace7294a5d64497a7bd85f7cbeb2/Source/MediaInfo/Audio/File_Ac3.cpp#L4256-L4261

Does not seem to match specs:

> The reserved_data_size_bits codeword indicates the value of reserved_data_size.
>
> The reserved_data_size indicates the length of the reserved_data element plus the padding element in bytes and can be calculated by the following equation:
>
>                             reserved_data_size = reserved_data_size_bits+ 1
>
> In this equation, the bits of the reserved_data_size_bits element are interpreted as an unsigned integer.

cjee21:
@JeromeMartinez There should be nothing else now. Almost complete parsing of TrueHD and OAMD now. Also CRC and parity computations for TrueHD. At the moment, only two of the CRC/parity are used, the others are only computed during trace. All CRC/parity results are only shown in trace. I did not do anything regarding the items mentioned in the above two posts.

JeromeMartinez:
> Not related to this PR but noticed that for raw TrueHD stream, even with full parse, there is no duration and bitrate in output.

It was not implemented for full parse, but should be easy, I'll give it a look.
For non full parse, there is no timestamp and it is VBR so not possible to get it.

cjee21:
Malformed EMDF E-AC3 samples that can crash Dolby's official decoders (CVE-2025-54957) may be useful for testing and can be found here:
https://project-zero.issues.chromium.org/issues/428075495

cjee21:
This PR is just for expanding MediaTrace and getting more info from TrueHD and DRC info from some JOC as well as trying to improve the display when both TrueHD and Ac3 in same track (but still not well done yet). It should not affect which files are parsed or not so

> files which were not well parsed without this PR

No.

PR #2382, opened 2025-08-28
IAMF/FLAC/Opus: More parsing

cjee21:


JeromeMartinez:
> For some reason, MKV attachments do have JPEG trace in output but not FLAC pictures.

There is a complete review of the trace feature there avoiding such issue, but also simplifying it a lot without regression, but I don't have the time for that for the moment (need to have regression tests first and so on :( ).

Issue #2375, opened 2025-08-02
Fails to Compile Against Graphviz 13.x


bassmadrigal:
Graphviz 13.0, released in June, had the following listed in its [changelog](https://gitlab.com/graphviz/graphviz/-/blob/main/CHANGELOG.md?ref_type=heads#1300--2025-06-08):

- Breaking: gvRenderData takes its length parameter as a size_t *. The GVJ_t fields output_data_allocated and output_data_position have type size_t.

When graphviz was updated to 13.x in my distro's repo, it broke libmediainfo when compiled with graphviz support, which is a package I maintain for that repo. When trying to compile against the newer graphviz, the error is:

```
../../../Source/MediaInfo/Export/Export_Graph.cpp:160:60: error: cannot convert ‘unsigned int*’ to ‘size_t*’ {aka ‘long unsigned int*’}
  160 |         gvRenderData(Context, Graph, "svg:cairo", &Buffer, &Size);
      |                                                            ^~~~~
      |                                                            |
      |                                                            unsigned int*
In file included from ../../../Source/MediaInfo/Export/Export_Graph.cpp:35:
/usr/include/graphviz/gvc.h:99:49: note:   initializing argument 5 of ‘int gvRenderData(GVC_t*, graph_t*, const char*, char**, size_t*)’
   99 |                          char **result, size_t *length);
      |                                         ~~~~~~~~^~~~~~
../../../Source/MediaInfo/Export/Export_Graph.cpp:162:54: error: cannot convert ‘unsigned int*’ to ‘size_t*’ {aka ‘long unsigned int*’}
  162 |         gvRenderData(Context, Graph, "svg", &Buffer, &Size);
      |                                                      ^~~~~
      |                                                      |
      |                                                      unsigned int*
In file included from ../../../Source/MediaInfo/Export/Export_Graph.cpp:35:
/usr/include/graphviz/gvc.h:99:49: note:   initializing argument 5 of ‘int gvRenderData(GVC_t*, graph_t*, const char*, char**, size_t*)’
   99 |                          char **result, size_t *length);
      |                                         ~~~~~~~~^~~~~~
```

We are able to fix this locally with the following sed command:

    sed -i 's|unsigned int|size_t|' Source/MediaInfo/Export/Export_Graph*

However, this is not backwards compatible with graphviz 12.x and earlier, so it would seem the code would need some sort of if/then statement to determine the graphviz version and use the correct parameters, but that is beyond my programming expertise.

PR #2388, opened 2025-09-11
Fix Graphviz 13+ support: version detection and size_t switch

bassmadrigal:
Potential fix for #2375.

Seems to work locally, but I do not have a ton of experience with C++ and autotools.

### DIFF
--- a/Project/GNU/Library/configure.ac
+++ b/Project/GNU/Library/configure.ac
@@ -556,6 +556,38 @@ else
 fi
 
 dnl -------------------------------------------------------------------------
+dnl Graphviz version detection
+
+if test "$using_graphviz" != "no" -a "$using_graphviz" != ""; then
+	# Try to detect Graphviz version using pkg-config
+	if ${PKG_CONFIG:-pkg-config} --exists libgvc; then
+		GV_PKG_VERSION=$(${PKG_CONFIG:-pkg-config} --modversion libgvc)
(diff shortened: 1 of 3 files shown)


cjee21:
I am not very familiar with this part of the codes as well but what I know is the official distributions for Linux and Windows are built with `--with-graphviz=runtime` (https://github.com/MediaArea/MediaInfoLib/commit/1eecc6fdd6730955035385f7a2da7e6dde585a55) which it will load Graphviz during runtime when using graph mode if the .so or .dll are found else it will safely fail with indication that graph plugin is not found. So I wonder in this case how are the versions handled.

JeromeMartinez:
Thank you for working on it.
As @cjee21 indicated, there is still a problem with the dynamic load, there is a need to detect the version during dynamic loading and bind the right interface (old or new) accordingly.
If you can do that it would be great, else we let open this ticket until we have time for doing it.

cjee21:
Looks like one way to determine the version during dynamic loading (for Windows at least, don't know about Linux version) is to check if there is `xml_escape` function exported in `gvc.dll`. It is no longer exported starting version 13 (https://gitlab.com/graphviz/graphviz/-/merge_requests/4278/diffs?commit_id=f7a9642928d1543b4d0e4bbf3cba7c290dcf43c8#add2a648a46f612f78e218ebe0d988ca7ee2e36d_0_1).

JeromeMartinez:
https://github.com/MediaArea/MediaInfoLib/pull/2551 is merged, for next release.

PR #2415, opened 2025-10-01
CMake: Enable ZLIBSTATIC

cjee21:
Also set CMake version range to prevent deprecation warnings.

### DIFF
--- a/.github/workflows/MediaInfoLib_Checks.yml
+++ b/.github/workflows/MediaInfoLib_Checks.yml
@@ -3,10 +3,10 @@ name: Checks
 on: [push, pull_request]
 
 jobs:
-  Unix:
+  Build:
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
         environement: [gcc, utf8, minified, minimal, minimal-with-xml, cmake]
         exclude:
           - os: macos-latest
@@ -14,6 +14,16 @@ jobs:
           - os: macos-latest
             environement: minimal
           - os: macos-latest
+            environement: minimal-with-xml
+          - os: windows-latest
+            environement: gcc
+          - os: windows-latest
+            environement: utf8
+          - os: windows-latest
+            environement: minified
+          - os: windows-latest
+            environement: minimal
+          - os: windows-latest
             environement: minimal-with-xml
         include:
           - environement: gcc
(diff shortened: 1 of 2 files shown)


JeromeMartinez:
> Also set CMake version range to prevent deprecation warnings.

Shouldn't max version be 3.30? I read from doc that 3.31 introduces the deprecation warnings for <3.10.

cjee21:
> Shouldn't max version be 3.30? I read from doc that 3.31 introduces the deprecation warnings for <3.10.

I don't know much about CMake. I saw the warning so I just put the current version shipped with MSVC there and the warning is gone.

cjee21:
It appears that after changing 3.31 to 3.30, another warning has returned:

```
CMake Warning (dev) at CMakeLists.txt:456 (install):
  Policy CMP0177 is not set: install() DESTINATION paths are normalized.  Run
  "cmake --help-policy CMP0177" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.
This warning is for project developers.  Use -Wno-dev to suppress it.
```

JeromeMartinez:
> Now there is only one deprecation warning left from ZenLib that should be gone after https://github.com/MediaArea/ZenLib/pull/185

I am not keen on the max value but fair enough for the moment, until someone does fix the warnings.

Thanks.

cjee21:
What about like this? I renamed the jobs to Build and MSBuild. Also added a define to prevent unused zlib tests/examples from being built.

cjee21:
No more warnings remain. I think using 3.31 as max should be alright. zlib uses this version as max too.

JeromeMartinez:
> What about like this? I renamed the jobs to Build and MSBuild. 

Fair enough for the moment.

>  I think using 3.31 as max should be alright.

I like to have fixes instead of limiting the version but CMake is definitely need my priority, especially because we don't use it ourselves, so I let other people manage that.
(with the constraint that I want to keep compatibility with very old CMake version, so it does not help to "fix" the warnings :) )

JeromeMartinez:
> If I understand correctly it is not limiting but stating that it has been tested up to 3.31 so there won't be deprecation warnings and so on.

Right. Just that I would like to have a CMake script without warning for all CMake versions, but I guess it is not possible, or at least not easily. And CMake 3.5 "format" is still supporte by CMake 4+ so fine enough. Just a nice to have and definitely not the priority :).
(I try to manage your other PRs, which are a lot more important than CMake, by the end of the month)

PR #2448, opened 2025-10-07
Rust wrapper and example

cjee21:
Features
- Bindings generator and Rust wrapper in `Source\MediaInfoDLL\MediaInfoDLL-rs`
- Example in `Source\Example\HowToUse_Dll-rs`
- Raw bindings are automatically generated from MediaInfoLib header during build
- Rust wrapper is manually written to wrap raw bindings
- Building the example will automatically generate and build everything required
- The CI builds and test runs the Rust example using package generated from MSBuild job for Windows x64 and ARM64

Limitations
- It is only tested on Windows 11 x64 and ARM64 at the moment.
- Currently only a limited amount of API that is used by the example has manually written wrapper. The rest can be accessed via raw bindings.


cjee21:
@JeromeMartinez Noticed the official packages on MediaArea.net for x86/i386 does not contain `*.lib` files.

JeromeMartinez:
> is MediaInfo_int8u* Buffer in Open_Buffer_Continue supposed to be const? 

It is const internally, and the C++ interface has the const, and the C interface redirects to the C++ interface., so we don't write at it, the cast would be OK (not sure we can add the "cast" to the interface without breaking the ABI).

cjee21:
Running `test_inform()` and `test_unicode()` in parallel will randomly fail. Seems the option is affecting the other instance.

cjee21:
> Noticed the official packages on MediaArea.net for x86/i386 does not contain `*.lib` files.

Looks like it has been missing .lib files for years due to some copy-paste bug (3 x64 shouldn't be there):

https://github.com/MediaArea/MediaInfoLib/blob/62d2ad842f698f8db616297886830264912953be/Release/Release_DLL_Windows_i386.bat#L132-L133

PR #2458, opened 2025-10-13
APV: Initial support

cjee21:
Specification and samples: https://github.com/AcademySoftwareFoundation/openapv
More info: https://developer.android.com/about/versions/16/features#apv

Samples in MP4 can be generated with FFmpeg:

```
ffmpeg -f lavfi -i testsrc=d=10:s=3840x2160:r=60 -pix_fmt yuv422p10le -c:v liboapv -preset slow
```

or for `color_description_present_flag` = 1:

```
ffmpeg -f lavfi -i testsrc=d=10:s=3840x2160:r=60 -pix_fmt yuv422p10le -vf "setparams=color_primaries=bt709:color_trc=bt709:colorspace=bt709" -c:v liboapv -preset slow
```


cjee21:
Rebased + fixed some issues flagged by Cppcheck and vulnerabilities flagged by AI that I missed.

JeromeMartinez (changes requested):
`CtsMediaV2TestCases/pattern_dynamic_aspect_1280x720_30fps_30mbps_apv_10bit.mp4` has a weird behavior, lot of stuff (profile, width, bit depth...) are repeated 3x in MediaInfo output.

cjee21:
> `CtsMediaV2TestCases/pattern_dynamic_aspect_1280x720_30fps_30mbps_apv_10bit.mp4` has a weird behavior, lot of stuff (profile, width, bit depth...) are repeated 3x in MediaInfo output.

Isn't that because stream contains 3 types (SDR, HDR10, HLG) of streams concat together?

JeromeMartinez:
> Isn't that because stream contains 3 types (SDR, HDR10, HLG) of streams concat together?

color related stuff should not impact profile, width, bit depth, especially if they are same.
I didn't check yet the bitstream, I'll do it in a couple of hours.

cjee21:
> color related stuff should not impact profile, width, bit depth, especially if they are same. I didn't check yet the bitstream, I'll do it in a couple of hours.

If I remember correctly, right now the parser will collect the number of different streams then fill each one. So filling same info repeatedly if the stream has same dimensions etc.

But let's say there are 3 streams with different color but only one has different bit depth, if we only show 2 bits depths and 3 colours, how can we tell which stream has which bit depth?

JeromeMartinez:
I updated the PR with what I have in mind about this parser.
This is mostly subjective opinion about what I want.
I moved the tile part into the generic header parser.
MediaInfo does not currently support well multiple configurations for a stream, so I peek the first configuration. Multiple configuration support is for later, when I see how I should manage that in MediaInfo as a whole.
I prefer to show Format level on the same line as Format profile, the XML and JSON output split in 2 elements automatically. At least until I figure out how to manage a 1 line display without this hack.
An APV frame may be big and we don't need the content in tiles, I am hesitating to just skip the tiles part, at least when trace is not requested, in order to avoid to load the frame (at least for raw APV frames for the moment, and later I'll support to parse not fully loaded frames when it is in a container).

cjee21:
Looks okay to me. Now only remains the ITU T35 part after that parser is done.

JeromeMartinez (approved):

JeromeMartinez:
Let's go with that for the moment.
The T35 parser PR is still on my backlog, soon.

PR #2466, opened 2025-10-31
Improve alignment of text output for CJK languages

### LINKED ISSUE #1183: MediaInfo abnormal display in English for Japanese tranlsation
MediaInfo git-v25.07-6-gaeb9abbd0;

Image

cjee21:
Follow-up of https://github.com/MediaArea/MediaInfoLib/pull/2201
Resolves https://github.com/MediaArea/MediaInfo/issues/1183

### DIFF
--- a/Source/MediaInfo/MediaInfo_Inform.cpp
+++ b/Source/MediaInfo/MediaInfo_Inform.cpp
@@ -99,7 +99,7 @@ Ztring Xml_Name_Escape_0_7_78 (const Ztring &Name)
 }
 
 //---------------------------------------------------------------------------
-std::wstring ToFullWidth(const std::wstring& input)
+static std::wstring ToFullWidth(const std::wstring& input)
 {
     std::wstring result;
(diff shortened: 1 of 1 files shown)


cjee21:
> Do I interpret well

This patch is for optimal display in terminals (default/any font) and also in GUI (with correct monospace font for selected language). Result is as seen at https://github.com/MediaArea/MediaInfo/issues/1183#issuecomment-3473477353

> prefer to keep the previous behavior by default ... patch as an option

Up to you and the others. It does not matter much to me as I only use English and only have filenames/tags in CJK languages which isn't affected since those are on the right side of the colons.

What should the option be called and I am not sure how to add an option.

cjee21:
`std::wstring_convert<std::codecvt_utf8<wchar_t>> Converter;` part (that is already there before this PR) is not supported in newer C++ standards and cannot be built with MSVC (build error) without silencing warnings.

PR #2483, opened 2025-12-17
IAMF: Support IAMF v2.0.0 test files

cjee21:
Test files are generated by GitHub Actions in https://github.com/AOMediaCodec/iamf-tools


JeromeMartinez:
This PR is very independent from other parsers, no regression spotted, so could be merged, is it ready for review or should we still wait?

JeromeMartinez:
I missed it before merging but I see know that test_000134.iamf is not detected. This is the only one.

Issue #2492, opened 2026-01-17
Cover no longer shown for MKV?


cjee21:
No idea when it changed but previously (tested 25.10 official release) it shows 'Cover' fields for MKVs with attachment named 'cover' but now (current master branch) it doesn't and this causes showing cover in HTML view to be no longer possible for MKVs.

Before:

```
Writing application                      : mkvmerge v63.0.0 ('Everything') 64-bit
Writing library                          : libebml v1.4.2 + libmatroska v1.6.4
Cover                                    : Yes
Cover description                        : cover.jpg
Cover type                               : Cover
Attachments                              : cover.jpg
```

Now:

```
Writing application                      : mkvmerge v63.0.0 ('Everything') 64-bit
Writing library                          : libebml v1.4.2 + libmatroska v1.6.4
Attachments                              : cover.jpg
```

cjee21:
558a4d6bf29762ddcd7d1df30c3ad84fe011c72b - Cover OK
0ae0ce6b7e6977f6b2ea5b206b44d6de539b5a6f - Cover gone

JeromeMartinez:
>  is this intended?

No, it was a quick debug thing I let in my commit :(, I remove it soon.

PR #2493, opened 2026-01-18
Matroska: Add timecode label readout, fix

JeromeMartinez:
Debuging code in the [previous PR](https://github.com/MediaArea/MediaInfoLib/pull/2488), fix https://github.com/MediaArea/MediaInfoLib/issues/2492#issuecomment-3764116153


cjee21:
@JeromeMartinez I wonder if the content in the following if statement was intentionally removed? It is kind of strange that the content is removed but the if statements remain.

https://github.com/MediaArea/MediaInfoLib/blob/aeb02fa50920ecf2bf3c564716a30f1c5d9df1f0/Source/MediaInfo/Multiple/File_Mk.cpp#L2618-L2623

JeromeMartinez:
> I wonder if the content in the following if statement was intentionally removed?

Same issue, fixed.

PR #2511, opened 2026-02-02
AV1: Fixes and improvements

cjee21:
2 commits split from Dolby Vision PR since that PR is already complex enough and not likely to be merged soon along with new commits.

- Fix trace offset addresses for Get_leb128
  - Same as https://github.com/MediaArea/MediaInfoLib/commit/c47175271c795930b472007fa2c3c0778e940f1d for IAMF that is already merged.

- Add country and terminal provider code names for ITU-T T.35 metadata to trace
- Identify AOMedia ITU-T T.35 metadata
  - Only has AFGS1 under AOMedia terminal provider code (no sample files containing this so purely based on [spec](https://aomediacodec.github.io/afgs1-spec/))

- Parse metadata_timecode and metadata_scalability

- Fix and improve parsing

PS Might need a better way to handle ITU-T T.35 soon as it is now used in several places and soon for AV2 as well.


JeromeMartinez:
> PS Might need a better way to handle ITU-T T.35 soon as it is now used in several places and soon for AV2 as well.

Yes, I'll add a T35 dedicated parser before hadling AV2 in order to let you bind to the parser and avoid to duplicate more code.

PR #2514, opened 2026-02-05
Code quality improvements

cjee21:
Fix https://github.com/MediaArea/MediaInfo/issues/1233


Source/MediaInfo/Reader/Reader_libcurl.cpp
```diff
@@ -793,6 +811,7 @@ bool Reader_libcurl_HomeIsSet()
     #ifdef WINDOWS_UWP
         return false; //Environment is not aviable
     #else
+        #pragma warning(suppress : 4996) //'getenv': This function or variable may be unsafe.
```
JeromeMartinez: revert that after getenv for catching other lines after this one.

PR #2516, opened 2026-02-08
AV1: Support of Annex B streams

JeromeMartinez:


JeromeMartinez:
> Now I just thought, should the use of scalability be indicated maybe in Format Settings? 

Yes

> Also maybe you want to do something with the timecode metadata.

File with timecode?

PR #2526, opened 2026-02-09
AV1 & AV2: reduce false-positives

JeromeMartinez:
cc @cjee21 I limit the false-positive but also handle all valid files, I think.


cjee21:
Still many "Size is wrong" in Annex B test streams.

Commenting out the following appears to make it parse properly but I do not really understand this part.

```c++
else if (IsAnnexB) {
    obu_size = Element_TotalSize_Get();
}
```

JeromeMartinez:
> Suggestion

Right, my mistake, better with that.
Added a couple of other tests, not 0% false positives for AV2 in AV1 argon but not far from it, fair enough for the moment, and nearly 100% of AV1 argon is well detected, undetected ones seem to be very buggy so fine.

Issue #1471, opened 2021-12-14
Auro-3D audio detection


paxter:
Is it possible to add an Auro-3D audio codec detection to mediainfolib?
If needed, I can provide examples.

JeromeMartinez:
Samples would be a first step for basic detection.
For more, well... even before saying that priority will depends on the [level of interest](https://mediaarea.net/Services), we need to find specs...

paxter:
I provided you three short samples.

1. https://send.cm/drg0zjl1lb03 (Auro-3D only)
2. https://send.cm/96ojtjsfkftc (Atmos and Auro-3D)
3. https://send.cm/jhr28n584n69 (Atmos and Auro-4D)

The audio title will show you the used audio type.
Hope that helps. Let me know, if you need anything further.

JeromeMartinez:
I checked files and looks like that Auro-3D is hidden in DTS, without any spec available. As is, it is impossible to detect Auro-3D without big reverse engineering. Would not be done on free support.

paxter:
Hm ok. Bad news, but thanks for checking @JeromeMartinez.

ValZapod:
Samples are gone. So DTS:X can have 2 LFE, and Auro can have the eye of God channel right up your head. Yeah, whatever.

paxter:
If needed, I can reupload the samples at any time. Just let me know.

ValZapod:
In mp4 they can be even uploaded here on github.

paxter:
The files are mkv and the size is between 1-2 GB.

ValZapod:
Okay, the no need!

gphilipp:
@paxter would you mind sharing the samples again (links are dead)?

paxter:
@gphilipp sorry, I can't find them anymore.

jaloji:
@gphilipp @JeromeMartinez I've some Auro3D files if you want

JeromeMartinez:
>  I've some Auro3D files if you want

Still interested (but with same issue about specs)
jerome@mediaarea.net if is can not be shared publicly.

jaloji:
@JeromeMartinez Sent.
Its Auro3D on DTS-HDMA, I don't have Auro3D on PCM yet.

JeromeMartinez:
> Its Auro3D on DTS-HDMA,

I see no place in the file for encoded extra content.

>  on PCM yet.

[...]

PR #2531, opened 2026-02-16
Auro-3D: Detect in 24-bit PCM

almirus:
Hello! This is the first version of support for displaying the Auro-3D format with channel layout. It was implemented by me based on reverse engineering; my detector is available here: https://github.com/almirus/Orua-D3

Auro is embedded in the PCM stream, so it can be contained in FLAC and DTS HD MA, and those must be decoded first.

The documentation is [here](https://github.com/almirus/Orua-D3/tree/main/Doc), and the test files are [here](https://github.com/almirus/Orua-D3/blob/main/test_files/link.md).


JeromeMartinez:
Great to have that.
We could use the FFmpeg plugin for decoding, in the future.
We can manage project files.

I'll force-push an update with the updated project files.

JeromeMartinez:
> Currently (this PR), only WAV files and MKV files containing WAV are supported.

I understand, the compressed streams will require a decoder, a totally other story, for later.

cinema-ONE:
Offering something for the part that was deliberately deferred here — @JeromeMartinez's *"the compressed streams will require a decoder, a totally other story, for later"*. No suggestion that it belongs in this PR; this is for whenever that later comes.

I have a working decode path for **Auro-3D inside DTS-HD MA and Blu-ray LPCM**, using ffmpeg exactly as the FFmpeg-plugin route would, plus a corpus to validate it against. It reads `.m2ts` straight out of `BDMV/STREAM`, so no remux step.

**What might be worth knowing before that work starts:**

- **~2 seconds of decoded audio is enough.** Measured 0.13 s wall clock including the ffmpeg call. The decode requirement is real but the *duration* is small, which may make it cheaper than it sounded earlier in #1471.
- **Block size varies per stream** — 800, 960, 1000 and 1024 all occur, and 1000 *and* 1024 each appear at both 48 and 96 kHz. It does not follow the sample rate. Reading it from the header (as @almirus does) is the only safe route; anything hard-coded passes the common case and silently misses the rest.
- **The CRC does all the work on false positives.** Sync alone is not enough: one DTS:X control produced 5,149 chance 16-runs on its LFE. Every one fails the CRC. On my corpus, 400 of 400 candidate blocks validate on each Auro file and 0 on each control.

**The offer:** ground truth from an **Auro-3D-certified hardware decoder** (StormAudio ISP Elite MK2 over its TCP API) rather than from another software implementation. I've run @almirus's method against it — 7 of 7, both negatives included, reported in [#1471](https://github.com/MediaArea/MediaInfoLib/issues/1471#issuecomment-5380937366). Corpus is two Auro-3D demonstration Blu-rays swept track-by-track (175 streams), two commercial remuxes and a demo-trailer set, spanning 8.0, 9.1, 10.1, 11.1, 11.1 (7+4) and 13.1, at 48 and 96 kHz, in both DTS-HD MA and LPCM.

If a compressed-carrier implementation lands and you want it checked against real hardware before merging, say the word and I'll run whatever set is useful. Happy to supply per-stream expected results too.

For reference, my own port of @almirus's method (CRC, header parse, ADOL walk to opcode `0x1E`) is at https://github.com/cinema-ONE/auro3d-detector — Python, MIT, ffmpeg the only dependency. It exists to test the method, not to compete with this PR.

cjee21:
> The decode requirement is real but the _duration_ is small

Then will just be enabling DTS-HD MA decoder in MediaArea's FFmpeg build (size and patent licensing consideration) and implementing the decoding with FFmpeg and passing the decoded to the detector (code complexity; is temp file needed? etc.).

cinema-ONE:
@cjee21 Two of those I can answer with measurements rather than opinion.

### The lossless extension is mandatory — core-only gets you nothing

I wondered whether the DTS **core** would be enough, since that would change the size and licensing calculus considerably. It is not. Same 4-second window from the same DTS-HD MA track, decoded two ways:

```
ffmpeg ... -c:a pcm_s24le                    -> Auro-3D, carrier 5.1, layout Auro 11.1
ffmpeg ... -core_only 1 -c:a pcm_s24le       -> nothing; no block passes CRC
```

That is by construction rather than bad luck: the payload lives in the least-significant bits of the *lossless* reconstruction, and the lossy core does not preserve them. So for DTS-HD MA carriers it is **XLL or nothing** — there is no cheaper decoder that yields a partial answer. Whether that is worth the build size and the licensing is entirely your and @JeromeMartinez's call, but at least the technical half is settled: there is no middle option to trade down to.

Worth adding that **LPCM carriers need no decoder at all** beyond what MediaInfo already does, which is what this PR covers today.

### No temp file needed

Pipe it. ffmpeg writes WAV to stdout, the detector parses the header and reads samples straight from the buffer:

```
ffmpeg -v error -ss <n> -i <file> -map 0:<stream> -t 2 -c:a pcm_s32le -f wav -
```

That is what my implementation does — no intermediate file, no seek-back. Two further properties that keep it cheap:

- **Peak memory is the decoded window, and the window is small.** 2 seconds is enough (the marker repeats ~48×/second). Worst realistic case, 8 channels at 96 kHz, is **5.9 MB**; a 5.1 track at 48 kHz is about 1.5 MB.
- **You can stop early.** Detection needs only a handful of CRC-valid blocks; there is no need to decode to the end, or to hold more than the current window.

One caveat from experience: **read the channel count and sample rate from the decoder's WAV header, not from the container.** On Blu-ray `.m2ts`, ffprobe frequently reports `channels=0` because it cannot resolve the layout from the container alone — five of the streams in my corpus did exactly that, and trusting the container value would have produced a divide-by-zero or a wrong stride.

Happy to run any build against the hardware-labelled corpus if that would help de-risk the decision.

PR #2532, opened 2026-02-17
MXF: fixed infinite loop in malformed files

TehMrCrowley:
While processing malformed MXF files, we observed that MediaInfo enters an infinite loop, repeatedly re-reading the same byte range and never returning from the query operation. This results in the calling application blocking indefinitely.

The issue was reproduced using the current 25.10 master branch. After investigation, we identified the root cause and implemented a fix in our fork. 

In addition to resolving the infinite loop condition, we modified the query logic to return false when a malformed MXF structure is detected, allowing the caller to handle the error gracefully instead of hanging.

### DIFF
--- a/Source/MediaInfo/File__Analyze.cpp
+++ b/Source/MediaInfo/File__Analyze.cpp
@@ -2923,6 +2923,12 @@ bool File__Analyze::Data_Manage()
             GoTo(File_Offset+Buffer_Offset+Element_Offset);
         else
             Buffer_Offset+=(size_t)Element_Offset;
+        if (Element_Size == 0 && Element_Offset == 0)
+        {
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
An element size of 0 may appear in other formats, and the action (stopping everything) is too hard, I aim to find a less binary solution.
Could you share a file demonstrating this issue? If not possible in public, drop a ail to info@mediaarea.net

JeromeMartinez:
Looks like that one of the commits added last week now handle such case, the file you provided is now fine without this commit, please test master branch.

JeromeMartinez:
Closing at it was resolved another way, please reopen if issue remains.

PR #2541, opened 2026-03-04
Move T35 parsing code to its own parser

JeromeMartinez:


cjee21:
> Would be a bug, please open a dedicated PR as it is not related to this PR.

There is also another inconsistency which causes compilation to fail for x64, Debug when precomp is enabled.

Wow, compilation is super fast with precomp enabled. So all these years wasted all those compile time due to this bug?

I'll make a PR with the fixes.

JeromeMartinez:
> Wow, compilation is super fast with precomp enabled.  So all these years wasted all those compile time due to this bug?

Same! I was sure I was using it and I was disliking the slow speed despite precompiled header, argh now I see that I introduced a mistake there...

cjee21:
FYI the things depending on this PR:
- ITU-T T.35 metadata OBU parsing in AV2
- ITU-T T.35 metadata OBU parsing in IAMF
- Eclipsa Video (SMPTE ST 2094-50) support - ITU-T T.35 in both MP4 meta track and OBU in AV1/AV2 (maybe VP9 and others too)
- Dolby Vision RPU parsing in AV1 (open PR) and AV2 (no samples yet)

PR #2548, opened 2026-03-07
Curl: Add default UserAgent + Use new curl_off_t API in runtime mode too

### LINKED ISSUE #2546: Should set an user agent for HTTP requests
The [curl-related code](https://github.com/MediaArea/MediaInfoLib/blob/master/Source/MediaInfo/Reader/Reader_libcurl.cpp) in MediaInfoLib does not set an user-agent ~~(or is there an option at all?)~~.
The HTTP RFC says that clients *should* send an user-agent field. In fact some web server configurations apparently refuse user-agent-less requests.

I propose mediainfo set an user agent by default. Something simple like "`MediaInfoLib`" would suffice.

cjee21:
Resolves #2546 

Now if a user agent is not set with `MI.Option("File_Curl;UserAgent", "ExampleUserAgentString")`, the default user agent in the format of `MediaInfoLib/<version>` will be used.

Test:

Screenshot 2026-03-07 at 23-13-25 Webhook site - Test transform and automate Web requests and emails

### DIFF
--- a/Source/MediaInfo/Reader/Reader_libcurl.cpp
+++ b/Source/MediaInfo/Reader/Reader_libcurl.cpp
@@ -1112,6 +1112,11 @@ size_t Reader_libcurl::Format_Test_PerParser(MediaInfo_Internal* MI, const Strin
         Curl_Data->Time_Max=time(0)+(time_t)MI->Config.File_TimeToLive_Get();
     if (!MI->Config.File_Curl_Get(__T("UserAgent")).empty())
         curl_easy_setopt(Curl_Data->Curl, CURLOPT_USERAGENT, MI->Config.File_Curl_Get(__T("UserAgent")).To_Local().c_str());
+    else {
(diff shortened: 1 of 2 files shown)


cjee21:
Previously we migrated to new curl_off_t API for CURLINFO_CONTENT_LENGTH_DOWNLOAD if libcurl is linked at compile time and the headers are newer than 7.55.0 and support the new API in order to prevent the deprecation warnings of using the old API. Now we also prefer the new API instead of using deprecated one in runtime linking mode but with fallback to maintain support of libcurl < 7.55.0. This required updating the runtime include headers to 7.55.0. Doing so also reduced the casts and conversions.

cjee21:
Also changed typedef of curl_off_t in runtime include header to int64s from int64u as libcurl sources state that it must be 64-bits signed.

JeromeMartinez:
thanks!

PR #2554, opened 2026-03-12
CMake: Version compatibility fixes

cjee21:


JeromeMartinez:
I like the idea to move to "the following warning is treated as an error" but need to fix warnings first.

cjee21:
> I like the idea to move to "the following warning is treated as an error" but need to fix warnings first.

Which warning are you referring to?

JeromeMartinez:
You changed something in your force-push, no more error, but [the initial CI](https://github.com/MediaArea/MediaInfoLib/actions/runs/22995666464/job/66767125001) failed, I guess due to a flag inadvertently added.

cjee21:
> You changed something in your force-push, no more error, but [the initial CI](https://github.com/MediaArea/MediaInfoLib/actions/runs/22995666464/job/66767125001) failed, I guess due to a flag inadvertently added.

Ah that is accidentally removed the flag to disable the warnings that MSVC has for MediaInfoLib C++ codes.

---

Hmm... The select highest C++ logic does not work properly on < CMake 3.8 and wrongly says compiler does not support C++11 so I put it in a version check. But now it looks like it will not set/require C++11 for older CMake versions.

JeromeMartinez:
> But now it looks like it will not set/require C++11 for older CMake versions.

You forgot to force "target_compile_features(${target} PRIVATE cxx_std_11)" as a "else", I guess.

JeromeMartinez:
And CXX_STANDARD_REQUIRED ON

cjee21:
cxx_std_11 is the thing that is not supported in older CMake. I guess I should put back the old logic of forcing CXX_STANDARD to 11 when on older CMake.

JeromeMartinez:
IN_LIST is CMake 3.12.
Rather tthan the ifdef, please replace (also in other projects) and test

```
if("cxx_std_${std}" IN_LIST CMAKE_CXX_COMPILE_FEATURES)
```

by

```
foreach(std ${_known_standards})
    list(FIND CMAKE_CXX_COMPILE_FEATURES "cxx_std_${std}" _index)
    if(NOT _index EQUAL -1)
        target_compile_features(${target} PRIVATE cxx_std_${std})
        message(STATUS "'${target}' is using latest available C++ standard: C++${std}")
        return()
    endif()
endforeach()
```

JeromeMartinez:
> cxx_std_11 is the thing that is not supported in older CMake.

It is supported by CMake 3.1+, not the issue IMO.

cjee21:
> It is supported by CMake 3.1+, not the issue IMO.

Hmm... https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html says "Added in version 3.8" ? Or am I looking at wrong thing.

JeromeMartinez:
I was fooled by the first line of this page, so 2 issues IN_LIST (easy fix) and cxx_std_11 (need to switch back to older method).
I am reluctant to raise now the CMake minimum version, if it is "only" that to fix.

PR #2559, opened 2026-03-26
Fixes for gcc 4.8.5 compatibility.

disc-kuraudo:
Trying to compile for legacy CentOS7/RHEL7 which is still in extended support for the latter with the stock gcc 4.8.5 compiler.

C++ 11 support of the compiler looks pretty complete to me so it _should_ work with this code:
https://en.cppreference.com/w/cpp/compiler_support/11.html
(only "Compiler support for type traits" seems to have partial support and got there with gcc version 5)

I fixed two minor issues that caused the following errors:
```
$ build_libmediainfo.sh |& grep error -A1
../../../Source/MediaInfo/File__Analyze_Streams_Finish.cpp:425:1: error: braces around scalar initializer for type ‘const char*’
 }; 
--
../../../Source/MediaInfo/File__Analyze_Streams_Finish.cpp:2102:1: error: braces around scalar initializer for type ‘const char*’
 };
--
../../../Source/ThirdParty/fmt/format.h:4108:16: error: missing space between ‘""’ and suffix identifier
 constexpr auto operator""_a(const char* s, size_t) -> detail::udl_arg<char> {
```

However there are still issues with some operator overloading stuff:
```
$ build_libmediainfo.sh |& grep error -A1
../../../Source/MediaInfo/Audio/File_Iamf.cpp:221:76: error: no match for ‘operator/’ (operand types are ‘std::initializer_list<long long unsigned int>’ and ‘ZenLib::float64 {aka double}’)
             Fill(Stream_Audio, 0, Audio_Duration, SamplesCountPerSubstream / (static_cast<float64>(SamplingRate) / 1000), 0);
--
../../../Source/MediaInfo/Audio/File_Iamf.cpp:223:88: error: no match for ‘operator/’ (operand types are ‘std::initializer_list<long long unsigned int>’ and ‘ZenLib::float64 {aka double}’)
             Fill(Stream_Audio, 0, Audio_BitRate, File_Size / (SamplesCountPerSubstream / static_cast<float64>(SamplingRate)) * 8, 0);
--
../../../Source/MediaInfo/Image/File_Png.cpp:1008:75: error: no match for call to ‘(std::initializer_list<MediaInfoLib::File_Png::Decode_RawProfile(const char*, size_t, const string&)::__lambda0>) (char*&, long unsigned int, long unsigned int&)’
     auto data = HexStringToBytes(end, in_len - (end - in), expected_length);
--
../../../Source/MediaInfo/Image/File_GainMap.cpp:81:25: error: no match for ‘operator<’ (operand types are ‘int’ and ‘std::initializer_list<int>’)
     for (int ch = 0; ch < channel_count; ++ch) {
--
../../../Source/MediaInfo/Image/File_GainMap.cpp:108:29: error: no match for ‘operator<’ (operand types are ‘int’ and ‘std::initializer_list<int>’)
         for (int ch = 0; ch < channel_count; ++ch) {
--
../../../Source/MediaInfo/Tag/File_Exif.cpp:1703:35: error: no match for ‘operator<’ (operand types are ‘const std::initializer_list<const double>’ and ‘double’)
                 if (exposure_time < 0.25001 && exposure_time > 0) {
--
../../../Source/MediaInfo/Tag/File_Exif.cpp:1703:62: error: no match for ‘operator>’ (operand types are ‘const std::initializer_list<const double>’ and ‘int’)
                 if (exposure_time < 0.25001 && exposure_time > 0) {
--
../../../Source/MediaInfo/Tag/File_Exif.cpp:1704:88: error: no match for ‘operator/’ (operand types are ‘int’ and ‘const std::initializer_list<const double>’)
                     ShutterSpeed_Time = "1/" + std::to_string(static_cast<int>(round(1 / exposure_time)));
--
../../../Source/MediaInfo/Tag/File_Exif.cpp:1707:69: error: no matching function for call to ‘to_string(const std::initializer_list<const double>&)’
                     ShutterSpeed_Time = std::to_string(exposure_time);
```

Is this something that could be looked at for improving portability? I wouldn't know how to fix the rest.

### DIFF
--- a/Source/MediaInfo/File__Analyze_Streams_Finish.cpp
+++ b/Source/MediaInfo/File__Analyze_Streams_Finish.cpp
@@ -420,8 +420,8 @@ struct FindReplaceCompany_struct {
 };
 
 static const FindReplaceCompany_struct Model_Replace[] = {
-    { "Canon", Model_Replace_Canon, {} },
-    { "OpenCube", Model_Replace_OpenCube, {} },
+    { "Canon", Model_Replace_Canon, nullptr },
+    { "OpenCube", Model_Replace_OpenCube, nullptr },
 }; 
 
 //---------------------------------------------------------------------------
@@ -2098,7 +2098,7 @@ static const FindReplaceCompany_struct Model_Name[] = {
     { "Sony", Model_Name_Sony, "Xperia " },
     { "Sony Ericsson", Model_Name_Sony_Ericsson, "Xperia " },
     { "Samsung", Model_Name_Samsung, "Galaxy " },
-    { "Xiaomi", Model_Name_Xiaomi, {} },
+    { "Xiaomi", Model_Name_Xiaomi, nullptr },
 };
 
 //---------------------------------------------------------------------------


cjee21:
Looks like the early C++11 implementation does not correctly parse auto and braces. All those are not supposed to be `initializer_list`. I can change the auto to explicit types.

cjee21:
@disc-kuraudo please test if https://github.com/MediaArea/MediaInfoLib/pull/2560 resolves the incorrect `std::initializer_list` deductions.

disc-kuraudo:
@cjee21 
Yes, this compiles together with my changes.

cjee21:
Regarding

```
../../../Source/ThirdParty/fmt/format.h:4108:16: error: missing space between ‘""’ and suffix identifier
 constexpr auto operator""_a(const char* s, size_t) -> detail::udl_arg<char> {
```

it is from third party code that is reported before

```
https://github.com/fmtlib/fmt/issues/4265
```

Looks like your fix will break other compilers.

disc-kuraudo:
I see, clang++ for instance issues a warning:
```
In file included from ../../../Source/ThirdParty/fmt/format.cc:8:
In file included from ../../../Source/ThirdParty/fmt/format-inl.h:23:
../../../Source/ThirdParty/fmt/format.h:4108:27: warning: identifier '_a' preceded by whitespace in a literal operator declaration is deprecated [-Wdeprecated-literal-operator]
 4108 | constexpr auto operator"" _a(const char* s, size_t) -> detail::udl_arg<char> {
      |                ~~~~~~~~~~~^~
      |                operator""_a
```

Could think of a messy workaround with `#ifdefs` checking if it is GCC and if GCC is `>= 4.9.x`, is there a better way?

I could work this around with a simple `sed` command for the time being and get it compiled on my end:
```
$ sed -e "s/operator\"\"_a/operator\"\" _a/g" -i Source/ThirdParty/fmt/format.h
```

cjee21:
I think you should just use sed for that due to the following reasons:
- fmt is third party code and usually changes are avoided in MediaInfoLib to keep it identical to upstream.
- fmt chose to only have no space and dropped support for GCC < 4.9
- newer C++ standards actually specify that it should be without space
- so far no other compilers have issues with that part

disc-kuraudo:
Yeah, I'll just revert that change.

JeromeMartinez:
> Trying to compile for legacy CentOS7/RHEL7

We (main maintainers of this lib) thought that nobody uses that anymore and we stopped to compile on it, looks like that we were wrong :-p.

> I could work this around with a simple sed command for the time being and get it compiled on my end:

I understand that this is less practical but I don't feel confortable changing a third party lib for that, especially if it breaks other compilers.
I am fine with adapting our code for being compatible with an old compiler if it has not impact on other compilers.
@disc-kuraudo please squash the commits.

Do I understand well if I can say that with the PR from @cjee21 (I merged it) + this modified PR (without libfmt related fix), the only remaining compilation issue is in libfmt and you can manage it relatively easily with a regex before lauching the compilation?

disc-kuraudo:
Squashed the commits.

Since it's a compiler error and RH never seemed to care patching this in RHEL, there is no need to do an ugly fix in the code.

Yes compilation runs through without problems after issuing the sed command above.

JeromeMartinez:
@disc-kuraudo tiny change request after another review, micro-optimization, please replace `""` by `nullptr` (should work too with GCC 4.8.5).

PR #2569, opened 2026-04-26
MZ/PE: Improvements

cjee21:
Improve the PE parser:
- More organized
- More scalable for future expansion
- More integrity checks
- Fix static analysis issues

Support JSON data added by cargo-auditable containing Rust dependencies.
- Only show the presence in MediaInfo output while JSON raw uncompressed data is only visible in MediaTrace as we do not have JSON parsing.

### DIFF
--- a/Source/MediaInfo/Archive/File_Mz.cpp
+++ b/Source/MediaInfo/Archive/File_Mz.cpp
@@ -22,6 +22,7 @@
 
 //---------------------------------------------------------------------------
 #include "MediaInfo/Archive/File_Mz.h"
+#include <zlib.h>
 using namespace ZenLib;
 //---------------------------------------------------------------------------
 
@@ -209,6 +210,11 @@ void File_Mz::Read_Buffer_Continue()
         to_parse.erase(to_parse.find(parsing_state));
         Goto_Next();
         return;
+    case State::CargoAuditable:
+        Parse_CargoAuditable();
+        to_parse.erase(to_parse.find(parsing_state));
+        Goto_Next();
+        return;
     }
 
     //Parsing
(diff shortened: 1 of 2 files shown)


JeromeMartinez:
> as we do not have JSON parsing.

Planned for C2PA too.

Source/MediaInfo/Archive/File_Mz.cpp
```diff
@@ -687,11 +706,67 @@ bool File_Mz::Parse_StringFileInfo(int8u level) {
...
+        auto Buffer_Size = static_cast<size_t>(strm.total_out);
+        inflateEnd(&strm);
+        // End of zlib decompression routine
+
+        // Only show raw JSON string in MediaTrace for now as we do not have JSON parsing
+        Ztring JSON;
+        JSON.From_UTF8(Buffer, Buffer_Size);
+        Param("Cargo Auditable JSON Data", JSON);
+    }
+#endif
```
JeromeMartinez, line 767: Add something like
```suggestion
    else
#endif
    Skip_XX(Element_Size, "Cargo Auditable JSON Data");
```

JeromeMartinez: Forget it, uselss.

JeromeMartinez:
> Support JSON data added by cargo-auditable containing Rust dependencies.

Please provide a basic ans small sample file I could add to my test files.

cjee21:
> Please provide a basic ans small sample file I could add to my test files.

I can make a small sample later. It can be made with `cargo install cargo-auditable` then `cargo auditable build --release` any Rust app with dependencies.

cjee21:
> > Support JSON data added by cargo-auditable containing Rust dependencies.
> 
> Please provide a basic ans small sample file I could add to my test files.

Here is sample which is MediaInfo's Rust HowToUse_Dll example built with `cargo auditable build --release`.

[HowToUse_Dll-rs_cargo-auditable.zip](https://github.com/user-attachments/files/27099305/HowToUse_Dll-rs_cargo-auditable.zip)

PR #2574, opened 2026-04-27
Matroska: support track flags from RFC 9559 (SDH, audio description, original, commentary)

### LINKED ISSUE #772: Show the following matroska flags: hearing-impaired (SDH) (for subs), commentary (for audios)
Currently mediainfo doesn't report these flags (even with `--full`).

zigazajc007:
Implements the five track flags from [RFC 9559](https://www.rfc-editor.org/rfc/rfc9559#name-track-flags) that MediaInfo currently ignores: `HearingImpaired`, `VisualImpaired`, `TextDescriptions`, `Original`, `Commentary`.

Closes https://github.com/MediaArea/MediaInfo/issues/772

image

image

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mk.cpp
+++ b/Source/MediaInfo/Multiple/File_Mk.cpp
@@ -127,6 +127,39 @@ namespace MediaInfoLib
 //***************************************************************************
 
 //---------------------------------------------------------------------------
+enum service_kind
+{
+    Service_HearingImpaired,
+    Service_VisuallyImpaired,
+    Service_TextDescriptions,
+    Service_Original,
+    Service_Commentary,
+    Service_Max,
+    Service_Default = Service_Max, // They are extra, actually not service
+    Service_Forced,
+};
+
+//---------------------------------------------------------------------------
+const char*  Matroska_ServiceKind[] =
+{
+    "HI",
+    "VI",
+    "TD",
+    "O",
+    "C",
+};
+
+//---------------------------------------------------------------------------
+const char* Matroska_ServiceKind_String[] =
+{
+    "Hearing Impaired",
+    "Visually Impaired",
+    "Text Descriptions",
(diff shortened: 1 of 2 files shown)


JeromeMartinez (changes requested):
Thanks for the PR.
But it creates too many MediaInfo report lines, the Default/Forced display was a mistake, I prefer to use the style used in e.g. AC-3 with a unique "Service kind" line when relevant.
Default/Forced will go there too at some point.

I adapted the PR but it looks like that I don't have the rights to force-push this PR, so please replaced your commit by the one in https://github.com/JeromeMartinez/MediaInfoLib/tree/Matrosla_ServiceKind

zigazajc007:
Do you think we should already include `Default` and `Forced` flags in `Service kind`, even if that temporarily creates duplicates?

That gives projects a clear indicator that these flags are expected to only be under `Service kind` in the future, so they can start adapting now without being surprised when the standalone `Default` and `Forced` lines eventually disappear.

zigazajc007 (commented):
I’m sorry, I wasn’t sure how to properly accept your changes, so I did them manually.

JeromeMartinez:
@zigazajc007 I forgot that FlagDefault is 1 by default, I added a line in my branch, please add it to your branch and squash the commits in 1 unique commit.

> Do you think we should already include Default and Forced flags in Service kind

Theses flags are not descriptive and have a technical hint for the player, so I don't know yet how I would handle them, I prefer to keep that as is for now and see how I can manage all formats.

> even if that temporarily creates duplicates?

We usually keep old MediaInfo fields, we just don't show them in the default text report, as we try to not break the API without the need of an incompatible change.

JeromeMartinez:
@zigazajc007 please squash the commits in 1 unique commit, no need to keep different commits here.

JeromeMartinez (approved):

PR #2582, opened 2026-05-05
Opus: Fix data skip size

cjee21:
Now that integrity checks have improved, this mistake is seen in trace.


cjee21:
"AV1/AV2: Remove unused includes" is moved here from another PR that is taking long to merge since it is small change with no functional effect.

JeromeMartinez:
I am on reducing the backlog 🙈.

PR #2590, opened 2026-05-09
Fix missing const in buffer API

cjee21:
Should not break anything since it is C API. Passing non-const buffer to it will still work but now can pass const buffer to it without doing "dangerous" casting away of const and callers will know it is guaranteed not to modify the buffer.


JeromeMartinez:
In theory it should not change anything in the ABI, let's see...

PR #2591, opened 2026-05-09
Fix DLL MediaInfoList_Open options not passed to internal

### LINKED ISSUE #2589: MediaInfoList_Open fileoptions is not used with DLL API
Discovered while working on Rust wrapper.

In DLL interface,

```C++
size_t          __stdcall MediaInfoList_Open (void* Handle, const wchar_t* File, const MediaInfo_fileoptions_C Options)
{
    MANAGE_SIZE_T(  "Open",
                    MediaInfoList,
                    Open(File),
                    Debug+=", File=";Debug+=Ztring(File).To_UTF8();)
}
```

which expands to

```C++
size_t          __stdcall MediaInfoList_Open (void* Handle, const wchar_t* File, const MediaInfo_fileoptions_C Options)
{
    Critical.Enter(); mi_outputs::iterator MI_Output = MI_Outputs.find(Handle); const bool MI_Output_IsOk = MI_Output != MI_Outputs.end(); Critical.Leave(); if (Handle == 0 || !MI_Output_IsOk) {
        return 0;
    } try {
        return ((MediaInfoList*)Handle)->Open(File);
    }
    catch (...) {
        return (size_t)-1;
    }
}
```

`const MediaInfo_fileoptions_C Options` is ignored and not pass to internal.

cjee21:
Fix #2589

### DIFF
--- a/Source/MediaInfoDLL/MediaInfoDLL-rs/src/lib.rs
+++ b/Source/MediaInfoDLL/MediaInfoDLL-rs/src/lib.rs
@@ -813,6 +813,11 @@ mod test_mediainfo {
             MediaInfo_fileoptions_t::MediaInfo_FileOption_Nothing,
         );
         assert_eq!(ret, 2);
+        let ret = mi.open(
+            "target/test.png",
+            MediaInfo_fileoptions_t::MediaInfo_FileOption_CloseAll,
+        );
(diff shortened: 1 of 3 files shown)


cjee21:
In this case, removing const has no effect to callers as it is passed by value and is C API.

PR #2592, opened 2026-05-09
Android library as Android Archive (AAR)

cjee21:
As previously discussed at https://github.com/MediaArea/MediaInfoLib/pull/2370#issuecomment-4351405286

Not sure if this is wanted but I just make a PR here with a PoC of building MediaInfoLib Android library as an AAR. Using AAR is documented at https://developer.android.com/studio/projects/android-library and is used by other libraries such as [libsignal](https://repo1.maven.org/maven2/org/signal/libsignal-android/0.86.5/), [c2pa](https://github.com/contentauth/c2pa-android/releases) and [geckoview](https://maven.mozilla.org/maven2/?prefix=maven2/org/mozilla/geckoview/geckoview/150.0.20260506140522/). Prefab is documented at https://developer.android.com/build/native-dependencies. AARs can be published to repositories as well for direct use in gradle projects.

This PR implements both JNI (with Kotlin wrapper from official app included) and C interface in the AAR.


cjee21:
Oh, and to verify the authenticity of the `gradle-wrapper.jar` binary in this PR, can find its SHA-256 here: https://gradle.org/release-checksums/

JeromeMartinez:
> Not sure if this is wanted 

This is wanted, for sure! We try to publish everywhere, just that we need help when we don't know much, and your help is precious.
We add that to our backlog (we still need to understand and test it).

cjee21:
> We try to publish everywhere

The Rust wrapper can be published as a crate at crates.io too (there are some abandoned third-party mediainfo crates there) but the way I did it is probably not ready to be published as it depends on files from repo or the official DLL package. I do not have experience publishing crates too.

cjee21:
I added an example app to test and demonstrate using the AAR with both the Kotlin interface and the native C/C++/CMake interface. Demo app written in Jetpack Compose.

Screenshot_20260516_230041
Screenshot_20260516_230046
Screenshot_20260516_230052

JeromeMartinez:
> I added an example app 

Great, very appreciated.

cjee21:
Now with libcurl support!

Screenshot_20260518_011707

JeromeMartinez:
> Now with libcurl support!

Great!
Note: it would be also nice to have that with the Android UI (and elsewhere, yeah... Late on that).

eeease:
@cjee21 When I tried running `gradlew.bat build :mediainfolib:assembleRelease`, I got an exception trying to GET the required cURL dependency:
`Execution failed for task ':mediainfolib:checkReleaseAarMetadata'.

Could not resolve all files for configuration ':mediainfolib:releaseRuntimeClasspath'.
Could not resolve io.github.vvb2060.ndk:curl:8.18.0.
Required by:
project ':mediainfolib'
> Could not resolve io.github.vvb2060.ndk:curl:8.18.0.
> Could not get resource 'https://repo.maven.apache.org/maven2/io/github/vvb2060/ndk/curl/8.18.0/curl-8.18.0.pom'.
> Could not GET 'https://repo.maven.apache.org/maven2/io/github/vvb2060/ndk/curl/8.18.0/curl-8.18.0.pom'.
> Got SSL handshake exception during request. It might be caused by SSL misconfiguration
> (certificate_unknown) PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target`

Any thoughts?

cjee21:
Looks like your system fails (does not trust the security cert) to establish https connection to repo.maven.apache.org repository? It works fine here. The cert for that repo is from Let's Encrypt and looks fine. You can build without the last commit to exclude libcurl if you want to build without libcurl.

eeease:
Thanks, I'm actually _re_building from this commit because I would like to include libcurl.
It does look like a corp laptop cert issue.  I'll see what I can do.  Having libcurl would definitely be a bonus.

eeease:
Update:  Yes, that was definitely a corp'd laptop VPN issue.  I'm building now.  Sorry for coming here so quickly with the issue!

PR #2594, opened 2026-05-10
JPEG: improve support of future content embedded in GContainer, update

JeromeMartinez:
video is in an MP4 then referenced by GContainer, so order should be:

```
Video
Muxing mode                              : GContainer XMP / MPEG-4
Stream size                              : 3.26 MiB (39%)
Type                                     : MotionPhoto
```


cjee21:
@JeromeMartinez is it expected that for the ones that are parsed by MPEG4 parser, `Muxing mode` only shows `GContainer XMP` without `/ MPEG-4`?

JeromeMartinez:
> @JeromeMartinez is it expected that for the ones that are parsed by MPEG4 parser, Muxing mode only shows GContainer XMP without / MPEG-4?

No, and I see, I worked with only files without first image offset (it is somewhere... but where???) e.g. Samsung Motion Photo, and I missed to fill also when first image offset is present so MP4 is parsed, I am adding that.

JeromeMartinez:
@cjee21 by the way, I see now how it is for Samsung (based on a header after the JPEG data, and the size of the JPEG is enough for getting the offset, and/or SEFT trailer).
Also, I have to better display width/height when the HEIF grid is used + SEFT in HEIF.
Still lot to do... But today it is release day, so such improvement is for the next release.

PR #2621, opened 2026-05-18
Enable linking to libcurl on Android and building without MD5

cjee21:
Enables linking to find_package(curl) in lower case to a Curl package that provides curl::curl_static target.

Also, building with Reader_libcurl requires enabling SHA256 which will enable MEDIAINFO_HASH but File_Md5 is not defined if MD5 is disabled. This makes it possible to build with Curl support but without MD5.


Source/MediaInfo/MediaInfo_Config_MediaInfo.cpp
```diff
@@ -1849,12 +1849,14 @@ bool MediaInfo_Config_MediaInfo::File_Md5_Get ()
 #if MEDIAINFO_HASH
 void MediaInfo_Config_MediaInfo::File_Hash_Set (HashWrapper::HashFunctions Funtions)
 {
+    #if MEDIAINFO_MD5
```
JeromeMartinez: Could be 4 lines later, couldn't it?

cjee21: I thought it is no use without the lower part but I saw wrongly.

Issue #2610, opened 2026-05-16
src/ZenLib/Source/ZenLib/BitStream_Fast.h:198 Global Buffer Overflow in ZenLib::BitStream_Fast::Get4


sigdevel:
#### Description:
When processing a crafted MP4 sample with a crafted `meta/iloc` box, MediaInfo crashes in `ZenLib::BitStream_Fast::Get4()` at `src/ZenLib/Source/ZenLib/BitStream_Fast.h:198` with a `global-buffer-overflow`.

`Get4()` takes `HowMany`, the number of bits to read as an `int8u` and uses it in two places at the end of the function without checking that it fits within the declared bounds:

```cpp
ToReturn |= (LastByte >> (Buffer_Size % 8)) & Mask[NewBits];
return ToReturn & Mask[HowMany];
```

`Mask` is `const int32u[33]`, so indices 0–32 are valid. `File_Mpeg4::meta_iloc()` at `File_Mpeg4_Elements.cpp:2717` calls `Get4(80)` based on a field decoded from the crafted `iloc` box. This causes both `ToReturn = LastByte << NewBits` with `NewBits = 80` (undefined shift on a 32-bit type) and `Mask[80]` (out-of-bounds table read). UBSAN reports both violations; ASAN reports the `global-buffer-overflow` on the `Mask` access at line 198.

#### Output:

asan-build:
<details>
  <summary>show relevant excerpt - click to expand</summary>

```bash
/run/media/user/81c71df6-ca99-4d27-a7b6-55107e347080/mediainfo/src/MediaInfoLib/Project/CMake/../../../ZenLib/Source/ZenLib/BitStream_Fast.h:184:30: runtime error: shift exponent 80 is too large for 32-bit type 'int'                                                                                                
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /run/media/user/81c71df6-ca99-4d27-a7b6-55107e347080/mediainfo/src/MediaInfoLib/Project/CMake/../../../ZenLib/Source/ZenLib/BitStream_Fast.h:184:30 
/run/media/user/81c71df6-ca99-4d27-a7b6-55107e347080/mediainfo/src/MediaInfoLib/Project/CMake/../../../ZenLib/Source/ZenLib/BitStream_Fast.h:198:47: runtime error: index 80 out of bounds for type 'const int32u[33]' (aka 'const unsigned int[33]')                                                                   
[...]

PR #2636, opened 2026-06-03
MPEG4: Validate sizes in iloc before using

cjee21:
Prevents https://github.com/MediaArea/MediaInfoLib/issues/2610

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mpeg4_Elements.cpp
+++ b/Source/MediaInfo/Multiple/File_Mpeg4_Elements.cpp
@@ -2673,6 +2673,22 @@ void File_Mpeg4::meta_iloc()
         index_size=0;
     }
     Get_S2 (16, item_count,                                     "item_count");
+    if (offset_size != 0 && offset_size != 4 && offset_size != 8) {
+        Trusted_IsNot("Invalid offset_size"); // TODO: Conformance error
+        return;
+    }
+    if (length_size != 0 && length_size != 4 && length_size != 8) {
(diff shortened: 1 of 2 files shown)


cjee21:
I do not know if you want to do this as conformance error or how to do the conformance error so I left that as a todo.

Source/MediaInfo/Multiple/File_Mpeg4_Elements.cpp
```diff
@@ -2673,6 +2673,22 @@ void File_Mpeg4::meta_iloc()
         index_size=0;
     }
     Get_S2 (16, item_count,                                     "item_count");
+    if (offset_size != 0 && offset_size != 4 && offset_size != 8) {
```
JeromeMartinez: I don't see any reason to prevent other values, so please change to ` if (offset_size > 8)`  (x4)

JeromeMartinez (changes requested):

Source/MediaInfo/Multiple/File_Mpeg4_Elements.cpp
cjee21: I followed spec specified values. I'll change it.

JeromeMartinez: > I followed spec specified values.

Well, we can handle other values... I prefer to have it generic, thanks.

JeromeMartinez (approved):

PR #2638, opened 2026-06-05
ELF: Support RISC-V

cjee21:


cjee21:
Sample file: https://github.com/cjee21/MediaInfo/actions/runs/27019770896

The sample file is a RISC-V build of MediaInfo CLI and it runs on RISC-V as seen on that GitHub Actions run :)

Source/MediaInfo/Archive/File_Elf.cpp
```diff
@@ -222,8 +226,12 @@ void File_Elf::Read_Buffer_Continue()
         Fill(Stream_General, 0, General_Format, "ELF");
         if (type!=(int16u)-1)
             Fill(Stream_General, 0, General_Format_Profile, Elf_type(type));
-        if (machine!=(int16u)-1)
-            Fill(Stream_General, 0, General_Format_Profile, Elf_machine(machine));
+        if (machine!=(int16u)-1) {
+            string Format_Profile = Elf_machine(machine);
+            if (machine == 243)
+                Format_Profile += classs == 1 ? " 32-bit address space" : classs == 2 ? " 64-bit address space" : "";
```
JeromeMartinez: I think I prefer a shorter form without extra " address space" part.

cjee21: I followed what you added in MZ/PE.

cjee21: I removed " address space" part for both ELF and MZ.

PS: AV2 typo fixes PR should be reviewed first. That is just minor typo but breaks parsing of some files.

JeromeMartinez: > I removed " address space" part for both ELF and MZ.

I prefer like that, thank you.

JeromeMartinez:
> The sample file is a RISC-V build of MediaInfo CLI and it runs on RISC-V as seen on that GitHub Actions run :)

I am interested in the risc-v build, so we even if we don't publish something for it yet, we are sure to not miss something there.

cjee21:
> I am interested in the risc-v build, so we even if we don't publish something for it yet, we are sure to not miss something there.

I used https://riscv-runners.riseproject.dev/ for testing but it is very slow. Ubuntu repo has MediaInfo for many architectures: https://launchpad.net/ubuntu/resolute/+package/mediainfo

PR #2640, opened 2026-06-12
CI: VS2026 + Ubuntu 26.04 + macOS 26

cjee21:


cjee21:
So that we can catch more issues with newer compilers which we did.....

```
Source/MediaInfo/Multiple/File_Mk.h:616:20: error: first argument in call to 'memset' is a pointer to non-trivially copyable type 'MediaInfoLib::File_Mk::rawcookedtrack' [-Werror,-Wnontrivial-memcall]
  616 |             memset(this, 0x00, sizeof(rawcookedtrack));
      |                    ^
/Users/runner/work/MediaInfoLib/MediaInfoLib/Project/CMake/../../Source/MediaInfo/Multiple/File_Mk.h:616:20: note: explicitly cast the pointer to silence this warning
  616 |             memset(this, 0x00, sizeof(rawcookedtrack));
      |                    ^
      |                    (void*)
```

cjee21:
Seems shouldn't use memset for overwriting complex structs. Also if `RawcookedTrack_Data=rawcookedtrack();` is executed when existing already has Buffer, there may be memory leak due to old destructor not called when reassigning.

PR #2645, opened 2026-07-01
Matroska: preserve LanguageIETF priority regardless of element order

maghuro:
## Summary

Preserves a Matroska track's `LanguageIETF` value regardless of whether the legacy `Language` element appears before or after it.

`LanguageIETF` is now handled separately and takes precedence over `Language` for the current `TrackEntry`.

This applies consistently to video, audio and text tracks.

## Why

Previously, both elements used the same handler. When `LanguageIETF` appeared first, a later legacy `Language` element could overwrite a specific BCP 47 tag such as `en-US` with `eng` / `en`.

## Validation

- Confirmed the change is limited to the Matroska parser.
- Preserved the repository's CRLF line endings.
- `git diff --check` passes.

Related to MediaArea/MediaInfo#1295

### DIFF
--- a/Source/MediaInfo/Multiple/File_Mk.cpp
+++ b/Source/MediaInfo/Multiple/File_Mk.cpp
@@ -3562,6 +3562,7 @@ void File_Mk::Segment_Tracks_TrackEntry()
     TrackType=(int64u)-1;
     TrackNumber=(int64u)-1;
     AudioBitDepth=(int64u)-1;
+    TrackHasLanguageIETF=false;
     TrackVideoDisplayWidth=0;
     TrackVideoDisplayHeight=0;
     AvgBytesPerSec=0;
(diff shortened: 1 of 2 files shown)


maghuro:
I retested this against the current MediaInfo/MediaInfoLib master and can confirm that the issue is still present.

For this test I built MediaInfo using:

```text
MediaInfo:    1d67fd0c8ddff5496eef9df6dcea9b98771df056
MediaInfoLib: 8bfa658657da9e16470c9fb32035e0fa097c0112
ZenLib:       2ddc277fe7ecfcbfe45616bb9cd9e23079113ecd
zlib:         eaaf237c8cbc7310170c43202c6ec2cff64fff66
```

The test file contains:

```text
Language     = eng
LanguageIETF = en-US
```

`mkvmerge -J` reports:

```text
ID=1 | LANG=eng | IETF=en-US | TITLE=English (US) | Opus 5.1 @ 689 kbps
```

Current MediaInfoLib master reports:

```text
ID=2 | LANG=en | TITLE=English (US) | Opus 5.1 @ 689 kbps
```

I then rebuilt MediaInfo using exactly the same MediaInfo, MediaInfoLib base, ZenLib and zlib commits, with only the patch from PR #2645 applied on top of the current MediaInfoLib master.

The result becomes:

```text
ID=2 | LANG=en-US | TITLE=English (US) | Opus 5.1 @ 689 kbps
```

So this is a controlled A/B test against the current master: the issue is still reproducible on `8bfa658657da9e16470c9fb32035e0fa097c0112`, and applying PR #2645 alone fixes it.

The PR also still applies cleanly on top of the current MediaInfoLib master.

PR #2648, opened 2026-07-09
Fix trace for Extended XMP chunks

cjee21:
Regression due to https://github.com/MediaArea/MediaInfoLib/pull/2571. For extended XMP we need to handle trace in JPEG. If we do it in XMP, will end up showing full assembled XMP in the last chunk and the other chunks are completely not shown due to empty trace from XMP parser. So revert to previous and add NoTrace = true to disable trace from XMP parser to show all chunks and actual contents on each chunk in Trace.


JeromeMartinez:
that is a workaround, I definitely need to find the origin of the issue.

cjee21:
> that is a workaround, I definitely need to find the origin of the issue.

For why need `Element_Show`, somewhere here right? Mentioned previously in some PR...

https://github.com/MediaArea/MediaInfoLib/blob/a2b519ce9b0a6c2dc9ecd34340576272b2093b5c/Source/MediaInfo/File__Analyze.cpp#L942

PR #2653, opened 2026-08-04
Fix/dtsx real object count

### LINKED ISSUE #2385: DTS:X - Channel Information
I noticed 25.07 is detecting more my DTS:X content (Although some like Zoolander 2 detect as 22bit depth). Is they any chance channel reporting could be added? Currently Mediainfo reports 7.1 plus object but in 99% of cases this is incorrect as most Bluray sourced DTS:X is 7.1.4 fixed or a mixture of fixed and objects (eg. 7.1.4 + 3 objects or 7.1.4 + 1 object).

almirus:


cjee21:
Resolves https://github.com/MediaArea/MediaInfoLib/issues/2385?

almirus:
Unfortunately, no. Currently, MediaInfo always reports “objects” (dynamic) for DTS:X, even though there may actually be none at all. This MR adds reporting their count.

cjee21:
So the DTS:X Object Emulator demo does not actually have objects? I assume it uses fixed height channels which MediaInfo does not show. If we can extract the fixed height channels and objects separately, then I think it is better if MediaInfo output is changed to be similar to the case of Dolby Atmos where we use `Number of dynamic objects`, `Bed channel count` and `Bed channel configuration` fields to indicate the fixed channels and objects instead of putting objects in the main Channel layout field. That would resolve the mentioned issue too. From what I know, only some older Yamaha AVRs can display the fixed channel layout and objects. Other decoders like the one I have only show DTS:X with no channel/object info so cannot be used as reference to check.

almirus:
> So the DTS:X Object Emulator demo does not actually have objects

Yes! Real objects can be found in Trinnov videos and in very rare films like Ip Man 3 — Otherwise, it’s just marketing.

almirus:
I’ll prepare an MR to extract the number of bed channels from DTS:X.

JeromeMartinez:
Thanks for the PR.
It seems fine with the files I have, but I don't get what it does not solve the reported issue ticket, if you have the count of objects you know when there is no object, what do I miss?

cjee21:
@JeromeMartinez you miss the height/bed channels. With this PR it will show Objects (number of objects) instead of just Objects and remove Objects when there are no objects. But MediaInfo does not show the height channels present in the bed. So with this PR, the DTS:X demo for example will say DTS:X but with only 7.1 channel and no objects. But it obviously has height effects when played back, so where does the height sound come from? This part (the fixed height channels) should be shown.

JeromeMartinez:
> you miss the height/bed channels. 

I get it now.

> I’ll prepare an MR to extract the number of bed channels from DTS:X.

@almirus it would be great!
I have no comment to add beside Copilot remarks, please update the MR with it and I would merge.
Thank you for that work, it helps.

JeromeMartinez:
> channel-based 7.1 streams such as Underwater from being reported as 7.1.4 while preserving confirmed supplemental layouts.

Is such file can be shared?

My main issue now is that you use your own bitstream parser, which makes the port to MediaInfo easier for you but it prevents us to use [MediaTrace](https://mediaarea.net/MediaTrace) feature, so I'll remove `dtsx_bit_reader` for using our classic `Get_BS()` stuff, but it may create regressions so it would be better if I have as many files as I can.

almirus:
This is a custom track (apparently the only one) that contains DTS:X metadata but has neither height channels nor dynamic objects.  https://limewire.com/d/pSCHq#oMYsHTuw0M

Here are some more files for testing.
https://disk.yandex.ru/d/wWfQkWpN4PmPRg

here is my decoder https://touch-max.ru/zvuk/dts-iznutri-gde-v-dts-hd-ma-spryatany-vysotnye-kanaly-i-obekty#download

JeromeMartinez:
> Here are some more files for testing.

Thanks, it will help in avoiding regression in the future.

JeromeMartinez:
Checking how Dolby Atmos and AC-4 display objects, actually the channel layout should have only the static channels and `NumberOfDynamicObjects` field should have the count of channels. " Objects" on this line was a mistake.
It may be tricky to change, we can let it as is fir this PR and I update it later, if you don't feel motivated for changing that.

JeromeMartinez:
I have 
```
Channel layout                           : C L R LFE Lb Rb Lss Rss Tfl Tfr Tbl Tbr Objects (2)
```

So looks like https://github.com/MediaArea/MediaInfoLib/issues/2385 would be well fixed with it, great!

PR #2675, opened 2026-08-31
Fix two issues with opus channel display

puellapurpura:
Fixes 5.0 being incorrectly displayed as 4.1, and channel layout using the wrong order for the center channel.


JeromeMartinez:
Sample files please.

puellapurpura:
Any 5.0 opus file will do, here's one that's just 1 second of silence (in zip since github wont let me upload it as .opus) [opus5.0.zip](https://github.com/user-attachments/files/31853618/opus5.0.zip)

PR #2682, opened 2026-09-11
Fix infinite loop when seeking in malformed matroska files

### LINKED ISSUE #2662: Hang and memory leak with corrupt mkv video file
Running this video with MediaInfo results in (I suspect) an infinite loop where it just continues to consume memory until the process is killed manually.

To reproduce:
`MediaInfo.exe corrupt.mkv`

Tested with v25.04 and 26.05
Running on Windows 11

[corrupt.zip](https://github.com/user-attachments/files/31255817/corrupt.zip)

Erik-White:
Closes #2662 

Fixes an infinite loop when seeking in malformed matroska files by adding a simple guard condition.

A truncated version of the original file attached to the issue is now checked in to the test files repo so it runs with the regression tests: https://github.com/MediaArea/MediaArea-RegressionTestingFiles/pull/6


Erik-White:
I just realised that adding that file to the test set could potentially cause CI hangs until this fix is merged. I hope it doesn't cause any issues!

Erik-White:
Thanks for the quick response! Do you have any kind of estimate when there will be a new release with this fix coming out?

JeromeMartinez:
> Do you have any kind of estimate when there will be a new release with this fix coming out?

End of the month.

Erik-White:
Great, thank you

Issue #2687, opened 2026-09-18
Uninitialized Node::Multiple corrupts JSON output for streams without any displayed field


buzz:
`struct Node` (`Source/MediaInfo/OutputHelpers.h`) has a default constructor that leaves `Multiple` uninitialized, while every other constructor sets it:

```cpp
    bool Multiple;
    bool AlreadyCommented=false;

    Node()
    {
    }
    Node(const char* _Name) : Name(_Name), Multiple(false)
```

`MediaInfo_Inform.cpp:600` is the only user of it, once per stream:

```cpp
    Node* Track=new Node();
    Track->RawContent=Inform((stream_t)StreamKind, StreamPos, false).To_UTF8();
```

`To_JSON_Elements()` then branches on that member (lines 267, 281, 314). For a stream with no field to display `RawContent` is empty, the node falls into the generic path, and the stale value decides the shape of the output: zero gives `"":null`, non-zero takes the object branch and also emits the array brackets of the two `if (Multiple)` blocks, which closes the `"track"` array before the last track object is closed. The result is not valid JSON; the corrupted tail is under Reproduction.

Affected: 26.05 and current master. Only `--Output=JSON`, `To_XML()` does not read the member.

## Reproduction

[`empty-menu-track.mp4`](https://raw.githubusercontent.com/buzz/mediainfo.js/test-fixtures/empty-menu-track.mp4), an MP4 with one chapter whose title is empty ([generator](https://github.com/buzz/mediainfo.js/blob/test-fixtures/generate-empty-menu-track.sh)). The chapter becomes a Menu stream with nothing to display, and it is the last stream, so the damage lands at the end of the document.

Emscripten/WASM build of 26.05, tail of `Inform()` after `Open_Buffer_*`:

```
,{"@type":"Menu","@typeorder":"2","":{}]}]}
}
```

`JSON.parse` rejects it at the first `]` after `{}`. Native mediainfo 26.05 prints this for the same bytes and is valid:

```
,{"@type":"Menu","@typeorder":"2","":null}]}
}
```

Whether the byte is zero depends on how much heap got recycled before that node, which is why native looks correct and the WASM build does not. Recompiling the same 26.05 with just

```cpp
    Node() : Multiple(false)
```

turns the WASM output into the valid one, so this is the read and not something downstream.

## Context

Reported by a user of mediainfo.js as [buzz/mediainfo.js#159](https://github.com/buzz/mediainfo.js/issues/159).

PR #2688, opened 2026-09-18
Fix uninitialized Node::Multiple breaking the JSON output

buzz:
The default constructor left Multiple uninitialized while every other constructor sets it, and `MediaInfo_Inform.cpp` creates one such node per stream. When a stream has no field to display, a non-zero stale value makes `To_JSON_Elements()` emit a stray bracket pair and close the "track" array too early, so `--Output=JSON` is not valid JSON. `To_XML()` does not read the member.

Native builds read back 0, an Emscripten build of the same version does not, so the corruption depends on the heap layout.

Issue: #2687

https://github.com/buzz/mediainfo.js/issues/159


JeromeMartinez:
Thank you!

PR #2689, opened 2026-09-18
Fix duration calculation of some containerless h264 essence files.

MaxEliaserAWS:
This fixes a regression from commit 5271548. This commit was meant to fix the
logic for determining the duration of a containerless h264 essence file, which
involves walking the frames in decode order and working out frame reordering
offsets.

While that commit may have fixed some cases, it actually *regressed* one of our
files. In particular, there is a frame reordering offset `pic_order_cnt_Delta`
that seems to "leak" across IDR boundaries and incorrectly affects the
timestamps calculated for a subsequent GOP. Definitionally, an IDR is supposed
to reset all decoder state and one GOP should not be influencing the timing of
a subsequent GOP. Therefore, we fix this by not applying the delta at IDR
frames.

In the case of this file, it is 241 frames at 24000:1001 framerate ratio. This
calculates out to 10.0517s. Prior to the regression, MediaInfo reported the
correct value rounded to the whole millisecond:
```
$ mediainfo --full /tmp/blackout.264 | grep -i duration
Duration                                 : 10052
Duration                                 : 10 s 52 ms
[...]
```
After the regression, MediaInfo adds an extra frame duration (appx 41-42 ms) to
the duration:
```
$ mediainfo --full /tmp/blackout.264 | grep -i duration
Duration                                 : 10093
Duration                                 : 10 s 93 ms
[...]
```
Comparing with other tools like ffprobe, I am pretty confident that the original
value of 10052 is correct, and this PR restores that value while not removing
the other functionality added with commit 5271548.

I have generated this file for regression testing:
[blackout.264.zip](https://github.com/user-attachments/files/32398856/blackout.264.zip)
You are free to save this file and use it for ongoing testing or any other
purpose, share it publicly, etc.

The fix itself was relatively easy, but generating a reproduction file *without*
leaking our customers' video content was pretty complex to do, since the issue
is reliant on the *exact* frame reordering and prediction dependency
relationships. Ended up hacking our encoder to do all the same analysis it
normally does but substitute black pixels at the last step, in order to generate
this file.


JeromeMartinez:
I agree the IDR should reset everything, I think that the proposed patch is not the best but the code is currently ugly so I keep this patch as is.
Thank you for the sample file, it helps a lot because I don't have any other file triggering this issue, it will help to avoid to have this bug again.

